PR #3, opened 2017-08-23
Initial implementation of recursively creating models

stevk:
TestValues.cs creates an object which holds the values we want to make models with. It also includes a helper function which creates a power set (all possible combinations) of those values.

Program.cs leverages this object to iterate through creating the models.

Note that I've commented out the contents of Materials.cs. The functionality of that code has been moved into Common.cs and Program.cs, but I thought to keep it around for a little bit longer as example code until we've more firmly landed this feature. 

**Outstanding issues:**

- [ ] Only the SingleTriangle fuction in common.cs has been updated with the gltf.Materials constructor. The other model types will need to be updated before they are used.

- [ ] All test values other than those used by PBR and basic materials need to be added. Textures should be added next.

- [ ] The powerset function does not handle required parameters yet

- [ ] I'm not happy with the current file naming scheme. It has the potential of creating files with way too long of names, given a large enough parameter set.

- [ ] Bunches of if statements in program.cs isn't great. There is probably a better way to do this.

### DIFF
--- a/Source/Program.cs
+++ b/Source/Program.cs
@@ -17,51 +17,67 @@ namespace AssetGenerator
 
             foreach (var combo in combos)
             {
-                string name = makeTest.GenerateName(combo);
+                    string name = makeTest.GenerateName(combo);
 
-                var gltf = new Gltf
-                {
-                    Asset = new Asset
+                    var gltf = new Gltf
                     {
-                        Generator = "glTF Asset Generator " + name,
-                        Version = "2.0",
-                    }
-                };
-
-                var dataList = new List<Data>();
-
-                var geometryData = new Data(name + ".bin");
-                dataList.Add(geometryData);
-
-                Common.SingleTriangle(gltf, geometryData);
-
-                gltf.Materials = new[]
-                {
-                    new Material
-                    {
-                        PbrMetallicRoughness = new MaterialPbrMetallicRoughness
+                        Asset = new Asset
                         {
-                            // MAKE THIS USE THE NEW CLASS!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-                            BaseColorFactor = new[] { 1.0f, 0.0f, 0.0f, 0.0f },
+                            Generator = "glTF Asset Generator " + name,
+                            Version = "2.0",
+                        }
+                    };
+
+                    var dataList = new List<Data>();
+
+                    var geometryData = new Data(name + ".bin");
+                    dataList.Add(geometryData);
+
+                    Common.SingleTriangle(gltf, geometryData);
+
+                    gltf.Materials = new[]
+                    {
+                        new Material
+                        {
+                            PbrMetallicRoughness = new MaterialPbrMetallicRoughness
+                            {
+                            }
+                        }
+                    };
+
+                    foreach (Parameter param in combo)
+                    {
+                        if (param.name == "BaseColorFactor")
+                        {
+                            gltf.Materials[0].PbrMetallicRoughness.BaseColorFactor = param.value;
+                        }
+
+                        if (param.name == "MetallicFactor")
+                        {
+                            gltf.Materials[0].PbrMetallicRoughness.MetallicFactor = param.value;
+                        }
+                        if (param.name == "RoughnessFactor")
+                        {
+                            gltf.Materials[0].PbrMetallicRoughness.RoughnessFactor = param.value;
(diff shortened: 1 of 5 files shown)


Source/Program.cs
```diff
@@ -12,43 +12,99 @@ private static void Main(string[] args)
...
+                    if (makeTest.testArea == Tests.materials)
+                    {
+                        foreach (Parameter param in combo)
+                        {
+                            if (param.name == ParameterName.Name)
                             {
-                                data.Writer.Flush();
+                                gltf.Materials[0].Name = param.value;
+                            }
+                            if (param.name == ParameterName.EmissiveFactor)
```
kcoley: you can switch these `if` statements to `else if` to avoid unnecessary comparison checks here.

stevk: Hey, good catch. I'll put that in.

Source/TestValues.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        public Parameter[][] ParameterCombos()
+        {
+            var temp = PowerSet<Parameter>(parameters);
+            //TODO: Remove sets that duplicate binary entries for a single parameter (e.g. alphaMode)
+            //TODO: Remove sets that exclude a required parameter
+
+            return temp;
+        }
+
```
stevk: @kcoley I'm trying to change this function to use a list instead of array, to cut down on the number of times I would otherwise need to switch back and forth. Line 71 in particular is giving me issues. Advice?

stevk:
Last commit reduced the number of models from the current materials param set from 128 to 96 by removing some duplicate sets introduced by non-binary parameters (alphaMode in this case).

Total unique models being generated currently is 112.

Source/TestValues.cs
kcoley: Maybe something like this?

```c#
public static List<List<T>> PowerSet<T>(List<T> seq)
        {
            var powerSet = new List<List<T>>(1 << seq.Count);
            powerSet[0] = new List<T>(); // starting only with empty set
            for (int i = 0; i < seq.Count; i++)
            {
                var cur = seq[i];
                int count = 1 << i; // doubling list each time
                for (int j = 0; j < count; j++)
                {
                    var source = powerSet[j];
                    var destination = powerSet[count + j] = new List<T>(source.Count + 1);
                    for (int q = 0; q < source.Count; q++)
                        destination[q] = source[q];
                    destination[source.Count] = cur;
                }
            }
            return powerSet;
        }
```

stevk: I tried something similar. It compiles but encounters "out of range" exceptions on the 4th and 12th lines.

` powerSet[0] = new List<T>(); // starting only with empty set`
seems to be fixed by 
`powerSet.Add(new List<T>()); // starting only with empty set`


but I'm not sure how to do the same thing for 
`var destination = powerSet[count + j] = new List<T>(source.Count + 1);
`

kcoley: Ah, Lists don't initialize size like arrays, but rather initialize capacity, so my example shouldn't work.  I'll rework it.

stevk: Thanks. I ended up working around the problem, but I think that in the long run we'd be better off working with lists. We're doing a lot of resizing, and it is easier to understand what's going on.

stevk:
@kcoley I fixed the conflict by making a ImageAttribute array in the TestValues class.

kcoley:
Cool sounds good.

kcoley:
Btw, here is a List implementation of the PowerSet function for reference (keeping the original function signature):

```c#
        public static T[][] PowerSet<T>(T[] seq)
        {
            // Create and initialize List of List<T>s
            var powerSet = new List<List<T>>((1 << seq.Length));
            for (int i = 0; i < (1 << seq.Length); ++i)
            {
                powerSet.Add(new List<T>());
            }

            for (int i = 0; i < seq.Length; ++i)
            {
                var cur = seq[i];
                int count = 1 << i;
                for ( int j = 0; j < count; ++j)
                {
                    var source = powerSet[j];

                    // Create and initialize List of size source.Count + 1
                    var destination = powerSet[count + j] = new List<T>(source.Count + 1);
                    for (int q = 0; q < (source.Count + 1); ++q)
                        destination.Add(default(T));
                    
                    for (int q = 0; q < (source.Count); ++q)
                        destination[q] = source[q];
                   
                    destination[source.Count] = cur;
                }
            }
            return powerSet.Select(a => a.ToArray()).ToArray();
        }
```

PR #4, opened 2017-08-23
Incorporate logic for handling local images in project

kcoley:


AssetGeneratorExamples/Examples.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="gltf"></param>
+        /// <param name="dataList"></param>
+        [Asset("TriangleWithTexture"), Image("green.png")]
+        public static void createTriangleWithTexture(string name, Gltf gltf, List<Data> dataList)
+        {
+            var geometryData = new Data(name + ".bin");
+            dataList.Add(geometryData);
+            GLTFWrapper wrapper = Common.SingleTriangleMultipleUVSetsWrapper(gltf, geometryData);
```
stevk: This would make more sense as an "expected input" comment, instead of as an example project.

AssetGeneratorExamples/Program.cs
```diff
@@ -0,0 +1,89 @@
...
+                            System.Diagnostics.Debug.WriteLine(method.Name);
+
+                            method.Invoke(null, new object[] { assetAttribute.Name, gltf, dataList });
+
+                            var assetFolder = Path.Combine(executingAssemblyFolder, assetGroupAttribute.Folder, assetAttribute.Name);
+                            Directory.CreateDirectory(assetFolder);
+                            var imageFolder = Path.Combine(executingAssemblyFolder, "ImageDependencies");
+                            System.Diagnostics.Debug.WriteLine(imageFolder);
+
+
```
stevk: We'll probably need something like this for copying images in the program.cs of the main program.

kcoley:
Good point.  I'll include this as commented information in the main project, and incorporate the logic for copying images from a project folder:

- [ ] Include the example usage as commented information in the main AssetGenerator project
- [ ] Incoporate the logic for copying images from a project folder into the AssetGenerator project
- [ ] Remove the AssetGeneratorExample project.

stevk:
I've just pushed a change to my branch that splits the models into separate folders based on what test is being run, so you can see what I have in mind.

Source/AssetGenerator.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26403.7
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
```
stevk: Was this change necessary? I'm using VS 2017 over here anyway.

kcoley: Think this version change isn't necessary.  I'll kick this out before merging.

- [ ] Keep original version for Visual Studio for the project

kcoley:
Changing the name of this PR to more describe what it is now doing.

kcoley:
The Image attribute can be used now as long as images are uploaded to the ImageDependencies directory in the project.

PR #13, opened 2017-08-25
Unit Tests

kcoley:
Initial pass for Unit tests.
This is a separate project within the solution and will be used to verify gltfwrapper components (Vectors, Matrices, Quaternions, etc.)


kcoley:
@stevk, I don't think this will step on your stuff since it's within a separate project, but you can double check.

PR #20, opened 2017-08-28
Add missing material parameters for name, alphaCutoff and doubleSided

kcoley:
Add material parameters for name, alphaCutoff and double sided to GLTFMaterial wrapper.
Also switched emissiveFactor to Vector3 instead of Vector4 to conform to spec requirements.


stevk:
Thanks for the quick fix!

kcoley:
No prob!  Thanks for catching

PR #21, opened 2017-08-28
Switch over to using wrapper to build gtTL models

stevk:
The wrapper is now used for creating the glTF objects.

### DIFF
--- a/Source/Program.cs
+++ b/Source/Program.cs
@@ -49,33 +49,30 @@ namespace AssetGenerator
                     {
                         foreach (Parameter param in combo)
                         {
-                            //if (param.name == ParameterName.Name)
-                            //{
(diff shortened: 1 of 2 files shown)


Source/Program.cs
```diff
@@ -41,58 +42,60 @@ private static void Main(string[] args)
...
+                                mat.doubleSided = param.value;
                             }
                         }
 
-                        gltf.Meshes[0].Primitives[0].Material = 0;
+                        wrapper.scenes[0].meshes[0].meshPrimitives[0].material = mat;
+                        wrapper.buildGLTF(gltf, geometryData);
                     }
 
                     if (makeTest.testArea == Tests.pbrMetallicRoughness)
```
kcoley: This line could be an `else if`

PR #23, opened 2017-08-28
Add missing fields for wrapper types (for materials and textures)

kcoley:
Added the following missing fields to these types (for the materials and textures implementation for the wrapper)

**GLTFTexture**
-  name

**GLTFImage**
-  mimeType
-  name

**GLTFSampler**
-   name

**GLTFMesh**
-  name

### DIFF
--- a/Source/GLTFWrapper.cs
+++ b/Source/GLTFWrapper.cs
@@ -264,6 +264,11 @@ namespace AssetGenerator
             /// List of meshes in the scene
             /// </summary>
             public List<GLTFMesh> meshes;
+
+            /// <summary>
+            /// The user-defined name of the scene
+            /// </summary>
+            public string name;
             public GLTFScene()
             {
                 meshes = new List<GLTFMesh>();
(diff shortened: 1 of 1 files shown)


Source/GLTFWrapper.cs
```diff
@@ -558,6 +598,14 @@ public Image convertToImage()
...
                     Uri = uri
                 };
+                if (mimeType.HasValue)
+                {
+                    image.MimeType = mimeType.Value;
+                }
+                if (name != null)
+                {
+                    image.Name = name;
+                }
```
stevk: It looks like a static name is being used if none is set. This isn't a required field per the spec, so we want the option to not pass a value in for this attribute.

stevk: Same for other locations where name is being set on null.

kcoley: Good catch.  I think that was legacy from my original default value pass.  I'll fix those.

kcoley: Actually, I think I had already reverted the default values.  Here, I don't set the name on the Image unless the GLTFImage's name value is not null.

stevk: Right, that is a not exception. My bad.

Issue #14, opened 2017-08-25
materials needs texture based parameters set


stevk:
- [x] normalTexture
- [x] occlusionTexture
- [x] emissiveTexture
- [x] pbr-baseColorTexture
- [x] pbr-metallicRoughnessTexture

Issue #11, opened 2017-08-25
Models are named after their parameters. This is unwieldly, long, and doesn't scale well.


stevk:
We still want an easy way to determine what parameters were used in a model's creation. 

Putting them all into the file name runs the following risks:

- [x] File name could exceed windows required file lengths
- [x] File name is hard to read due to no separation character

stevk:
I think the solution to this is to:

- [ ] List the attributes used in the "asset" portion of the file, along with the generator and version.
- [ ] Name the file using a combination of the test name and the number of files created.

kcoley:
+1 on including the data within the glTF file, which can work in asset portion. 

Another idea would be to group the glTF files into subdirectories and include a text file in the root directory referring to which parameters are in which folder, if that can help with querying.

PR #29, opened 2017-08-29
Create models that use PBR materials textures

stevk:
The pbrMetallicRoughness test now includes both BaseColorTexture and MetallicRoughnessTexture attributes.

This starts to address issue #14 and a small amount of work was done on #11 

This change expands the number of models being created to a whopping 4721.


kcoley:
I tried running this, but got an System.IO.PathTooLongException.

It conked out on this asset file name:

> "BaseColorFactor-MetallicFactor-RoughnessFactor-BaseColorTexture-Source-Sampler-TexCoord-Name-MetallicRoughnessTexture-Source-Sampler-TexCoord-Name"

>

kcoley:
Looks like it's because my running directory folder path is 124 chars, and the file name is 149 chars, which exceeds the 260 limit.  I can try running this in a different directory.

kcoley:
Runs fine in a shorter directory path.  Since it's not heavily being used by many people, it's probably okay for now.

stevk:
I was worried that we'd run into that issue #11 . I'll work on a fix in my next PR, since it is only going to get worse as we go along.

PR #35, opened 2017-08-30
Implemented material textures

stevk:
Added the remaining material textures to the material test:
-  normalTexture
-  occlusionTexture
-  emissiveTexture

This PR fixes issue #14 

We're now creating 43,506 models.


kcoley:
Do the gltf files generated in the materials folder have samplers and textures for the images?

stevk:
Not every combination has a sampler and source.

 I'm a bit on the fence about the ones that don't have a source, as that is a bit closer to testing negative cases. The source-less models should still load, but no texture will be displayed since there is no indexed texture for it to fall back on.

I'm also considering commenting out the 'name' attribute, due to the sheer number of models this test is creating.

kcoley:
Gotcha.  If there are samplers and sources in some of the combinations, then this should be fine.  I wrote a utility script to help me find files with samplers in them.

PR #59, opened 2017-09-07
Materials and Samplers test models

stevk:
Merging develop back into master now that we've stabilized. 

Currently we're creating 42 models, which represent the majority of what we're interested in testing in the materials and textures attributes.

Of note, we're missing tests for different texture file types and URI locations.


stevk:
These things need to be done before we merge into master:

- [ ] Alpha is set as 0 for the basecolorfactor. This should be higher so the model will be visible still.
- [ ] As part of the log, create a MD data table to show model attributes in a grid.
- [ ] Remove the Materials.cs file.
- [ ] Add a readme to introduce/explain project.
- [ ] bghgary's nuget package needs to be easily publicly available.
- [ ] Get extras object to hold assets name (Stretch goal).

stevk:
- [ ] #81 "Sampler test models are only using 3 UV coords"
- [ ] #80 "Sampler texture is not wrapped in a way that will result in a visible change with different WrapT and WrapS settings "

PR #60, opened 2017-09-07
changed default basecolorfactor to not create transparent models

stevk:
Basecolorfactor had alpha set to 0, which would have made models not render.

### DIFF
--- a/Source/TestValues.cs
+++ b/Source/TestValues.cs
@@ -77,7 +77,7 @@ namespace AssetGenerator
                         };
                         parameters = new List<Parameter>
                         {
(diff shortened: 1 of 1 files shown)


stevk:
We do need the alpha factor set to something less than 1, or else the non-default alpha modes won't do anything. I'll up it to 8 though. That should still be noticeable without being difficult.

 I probably need to add a couple more combos here, with alpha factor set to 1, and less than 1 for each mode.

stevk:
I assume that only PbrMetallicRoughness_0 is affected?

kcoley:
Yeah that one is probably the most noticeable.  Probably not a big deal for now.

PR #82, opened 2017-09-12
Remove Repeat WrapS and WrapT in Sampler test

stevk:
WrapS and WrapT use Repeat as a default value. The value is just thrown away if we try to set it, so there is no point in making a model just for this value.


stevk:
https://gist.github.com/stevk/dc504f6f8106941af42b083f79a73120

PR #85, opened 2017-09-12
flipped plane model

kcoley:
This flips the plane model so that negative z is forward.

### DIFF
--- a/Source/Common.cs
+++ b/Source/Common.cs
@@ -459,30 +459,30 @@ namespace AssetGenerator
             List<Vector3> planePositions = new List<Vector3>()
             {
                 new Vector3( 0.0f, 0.0f, 0.0f),
-                new Vector3( 0.0f, 1.0f, 0.0f),
+                new Vector3( -1.0f, 0.0f, 0.0f),
+                new Vector3(0.0f, 1.0f, 0.0f),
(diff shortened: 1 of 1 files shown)


Source/Common.cs
```diff
@@ -459,30 +459,30 @@ public static Runtime.GLTF SinglePlaneWrapper(Gltf gltf, Data geometryData)
...
-                    new Vector2(1.0f, 1.0f),
-                    new Vector2(1.0f, 0.0f),
-                    new Vector2(0.0f, 1.0f),
                     new Vector2(0.0f, 1.0f),
+                    new Vector2(1.0f, 1.0f),
+                    new Vector2(0.0f, 0.0f),
+                    new Vector2(1.0f, 1.0f),
                     new Vector2(1.0f, 0.0f),
                     new Vector2(0.0f, 0.0f)
                 },
```
stevk: Change the UV to this
```
new Vector2(2.0f, 2.0f),
new Vector2(-1.0f, 2.0f),
new Vector2(2.0f, -1.0f),
new Vector2(-1.0f, 2.0f),
new Vector2(-1.0f, -1.0f),
new Vector2(2.0f, -1.0f)
```

PR #89, opened 2017-09-13
morphtarget support

kcoley:
This adds morph target support to the runtime layer.  IT uses multiple mesh primitives as containers for other morph targets, assuming the positions and normals are associated with the same root mesh primitive

### DIFF
--- a/Source/Runtime/MeshPrimitive.cs
+++ b/Source/Runtime/MeshPrimitive.cs
@@ -14,6 +14,7 @@ namespace AssetGenerator.Runtime
     /// </summary>
     public class MeshPrimitive
     {
+        
         /// <summary>
         /// Material for the mesh primitive
         /// </summary>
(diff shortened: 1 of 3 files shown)


Source/Runtime/MeshPrimitive.cs
```diff
@@ -213,11 +220,11 @@ private glTFLoader.Schema.Accessor CreateAccessor(int bufferview_index, int? byt
...
-            if (min.Count() > 0)
+            if (min != null && min.Count() > 0)
             {
                 accessor.Min = min;
             };
-            if (max.Count() > 0)
+            if (max != null && max.Count() > 0)
             {
                 accessor.Max = max;
             }
```
stevk: I *think* that `.Any()` is supposed to be more performant than `.Count() > 0`

PR #91, opened 2017-09-13
Generated Assets v0.1

kcoley:
Changed name Sampler to Samplers to be consistent with Materials and changed .md files to README.md.  This should allow for the README.md to display on viewing the models.

This is a one-off generated commit of these assets to be tied with the master release as opposed to develop.


kcoley:
Btw, I changed Sampler and README.md manually, so this change isn't in the actual code for this pass.

stevk:
Per Gary's earlier suggestion, I think it would be better to make a branch off of master that is specifically for holding the pre-generated models, and then make PRs and Merges into that branch. That way if we do something that would change an already established model, we can see it during the PR.

1) Make a branch off of master [GeneratedAssets]
2) Check in a commit that removes the source code from [GeneratedAssets]
3) Make a branch off of [GeneratedAssets] in order to make the commit and PR containing the generated models
4) Repeat step 3 as we reach milestones on generating models

stevk:
Also, is naming the md logs README required for them to be displayed in git? If so, that change should just be checked into code instead of doing it manually. 

Same with the sampler -> samplers change

kcoley:
This PR can be used to follow steps 1 and 2 and have it be a PR to a new branch generated for 3.  I will remove the code in this branch.

Github displays the .md files when they are named README.md.  The Samplers and README.md name changes should definitely be made to the code-base.

kcoley:
A new branch has been created called generated-assets.  A PR with this commit has been made to that branch, so closing this one out.

PR #93, opened 2017-09-13
generated assets v0.1

kcoley:


kcoley:
MIT License has been re-added

PR #95, opened 2017-09-13
referenced generated-assets branch in README

kcoley:


kcoley:
branch URL didn't show up the way I expected.  Fixing that...

kcoley:
using absolute path to point to generated-assets branch now.

PR #108, opened 2017-09-15
convert all vectors to generics to be able to write byte and ushort to binary

kcoley:


Source/Runtime/MeshPrimitive.cs
```diff
@@ -23,27 +23,26 @@ public class MeshPrimitive
...
         /// </summary>
-        public List<Vector3> Colors { get; set; }
+        public List<Vector3<float>> Colors { get; set; }
 
         /// <summary>
-        /// List of texture coordinate sets (as lists of Vector2) 
+        /// List of texture coordinate sets (as lists of Vector2<float>) 
         /// </summary>
-        public List<List<Vector2>> TextureCoordSets { get; set; }
+        public List<List<Vector2<float>>> TextureCoordSets { get; set; }
```
stevk: Will declaring `Colors` or `TextureCoordSets` as <float> here cause issues when the texcoord or color is set as a <byte> or <ushort>?

kcoley: Yeah, this would cause issues.  I'll add byte and ushort

kcoley:
Going to instead abstract the ubyte and float in the runtime layer instead of converting vectors to generics.

PR #110, opened 2017-09-15
convert to ushort or byte when normalized booleans enabled for mesh primitive

kcoley:
`Runtime.MeshPrimitive` has four properties for controlling how the accessors store binary data for Colors and Texture Coordinates.

Setting `NormalizeColorCoords` to `true` and setting the `ColorComponentType` will use this value when creating the accessor and write in the appropriate format.  The same happens for `NormalizeTextureCoords` and `TextureCoordsComponentType`

### DIFF
--- a/Source/Runtime/MeshPrimitive.cs
+++ b/Source/Runtime/MeshPrimitive.cs
@@ -14,7 +14,12 @@ namespace AssetGenerator.Runtime
     /// </summary>
     public class MeshPrimitive
     {
-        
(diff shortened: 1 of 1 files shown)


kcoley:
Different implementation this time, which uses two enums to set the Color and TextureCoords Accessor Modes:

`ColorAccessorMode` and `TextureCoordsAccessorMode` can be defined to determine which mode to use for accessors.

The modes are `FLOAT`, `NORMALIZED_USHORT` AND `NORMALIZED_UBYTE`.  By default, `FLOAT` is used.

Issue #111, opened 2017-09-15
MeshPrimitive.Tangents requires Vec3 instead of Vec4


stevk:
Spec requires Vev4 for Tangent.

> XYZW vertex tangents where w is a sign value (-1 or +1) indicating handedness of the tangent basis

kcoley:
PR created #115

kcoley:
Merged.

PR #112, opened 2017-09-15
Implement Primitive Attributes Test

stevk:
Implemented a test for Primitive attributes.


stevk:
This code isn't completely done yet, but most of the work left is review and tweaks.

- [ ] Need a fix for #111 
- [ ] Verify models are working/rendering as expected
- [ ] Double check log output is formatted correctly

stevk:
https://gist.github.com/stevk/caf93075a584d66a0e5f848c2e94798a

stevk:
I'm thinking that normal/tangent doesn't make sense without a texture/uv set (e.g. basecolortexture). Is this the case?

kcoley:
It could be possible to apply a constant basecolorfactor and use the normal/tangent.  Though without either basecolorfactor or basecolortexture, I agree it would not make much sense.

stevk:
This PR should be ready to complete.

kcoley:
I may need to add in the byte padding update so the numbers should be correct for Model 11 (should show 8 bytes instead of 6).  I'll set that PR up

kcoley:
Can you add two more tests like PrimitiveAttributes1, but using all UNSIGNED_BYTE and all UNSIGNED_SHORT for both the TEXCOORDs and COLOR?

stevk:
Yup, I can do that.

stevk:
combinations of vec3 vs vec4 with that as well?

For example:
Model 1:
TexCoord0_Short
TexCoord1_Short
Color_Vec3_Short

Model 2:
TexCoord0_Short
TexCoord1_Short
Color_Vec4_Short

stevk:
Combos added and gist updated

stevk:
- [ ] Fix the md label "Normal Tangent" to be more human readable

stevk:
Last commit changed the MD for materials as well.

https://gist.github.com/stevk/caf93075a584d66a0e5f848c2e94798a

https://gist.github.com/stevk/6586744c3b3931f73494c4f56f20e010

stevk:
Updated PrimitiveAttribute gist
https://gist.github.com/stevk/caf93075a584d66a0e5f848c2e94798a

PR #123, opened 2017-09-19
Fix the color implementation in MeshPrimitive

kcoley:
**Changes:**

The `Color` attribute of `Runtime.MeshPrimitive` is now defined as `Vector4` instead of `Vector3`.

The color enum has been renamed from `ColorAccessorModes` to `ColorAccessorModeEnum` and has a Flag Attribute applied.  

i.e. To specify Float Vec4:
`meshPrimitive.ColorAccessorMode = Runtime.MeshPrimitive.ColorAccessorModeEnum.FLOAT | Runtime.MeshPrimitive.ColorAccessorModeEnum.VEC4;`

And to specify ubyte Vec3:
`meshPrimitive.ColorAccessorMode = Runtime.MeshPrimitive.ColorAccessorModeEnum.NORMALIZED_UBYTE | Runtime.MeshPrimitive.ColorAccessorModeEnum.VEC3;`

`TextureCoordsAccessorModes` has also been renamed to `TextureCoordsAccessorModeEnum` to be consistent with the Color enum.

A unit test has been added to test the flags of the `ColorAccessorEnum`

### DIFF
--- a/Source/AssetGeneratorTests/Runtime/MeshPrimitiveTests.cs
+++ b/Source/AssetGeneratorTests/Runtime/MeshPrimitiveTests.cs
@@ -147,6 +147,21 @@ namespace AssetGenerator.Runtime.Tests
             Assert.IsTrue(m.Primitives[0].Targets.Count() > 0);
             Assert.IsTrue(m.Weights.Count() > 0);
         }
+        [TestMethod()]
+        public void ColorAttributeEnumTest()
+        {
+            MeshPrimitive meshPrimitive = new MeshPrimitive();
+
+            meshPrimitive.ColorAccessorMode = MeshPrimitive.ColorAccessorModeEnum.FLOAT | MeshPrimitive.ColorAccessorModeEnum.VEC3;
+            Assert.AreEqual(meshPrimitive.ColorAccessorMode, MeshPrimitive.ColorAccessorModeEnum.FLOAT | MeshPrimitive.ColorAccessorModeEnum.VEC3);
+
+            meshPrimitive.ColorAccessorMode = MeshPrimitive.ColorAccessorModeEnum.NORMALIZED_UBYTE | MeshPrimitive.ColorAccessorModeEnum.VEC4;
+            Assert.AreEqual(meshPrimitive.ColorAccessorMode, MeshPrimitive.ColorAccessorModeEnum.NORMALIZED_UBYTE | MeshPrimitive.ColorAccessorModeEnum.VEC4);
+
+            
+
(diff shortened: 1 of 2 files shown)


Source/AssetGeneratorTests/Runtime/MeshPrimitiveTests.cs
```diff
@@ -125,7 +124,7 @@ public void GetMorphTargetsTest()
             glTFLoader.Schema.Buffer buffer = new glTFLoader.Schema.Buffer();
             Data geometryData = new Data("test.bin");
             int buffer_index = 0;
-            
+
```
stevk: Extra linebreak?

Source/AssetGeneratorTests/Runtime/MeshPrimitiveTests.cs
```diff
@@ -148,5 +147,21 @@ public void GetMorphTargetsTest()
...
+            meshPrimitive.ColorAccessorMode = MeshPrimitive.ColorAccessorModeEnum.NORMALIZED_UBYTE | MeshPrimitive.ColorAccessorModeEnum.VEC4;
+            Assert.AreEqual(meshPrimitive.ColorAccessorMode, MeshPrimitive.ColorAccessorModeEnum.NORMALIZED_UBYTE | MeshPrimitive.ColorAccessorModeEnum.VEC4);
+
+            
+
+
+        }
+
     }
 }
```
stevk: Nit: Missing a linebreak at the end of the file

Source/Runtime/MeshPrimitive.cs
```diff
@@ -49,8 +47,7 @@ public enum TextureCoordsAccessorModes { FLOAT, NORMALIZED_USHORT, NORMALIZED_UB
         /// <summary>
         /// List of colors for the mesh primitive
         /// </summary>
-        public List<Vector3> Colors { get; set; }
-
+        public List<Vector4> Colors { get; set; }
```
stevk: On second though, I'm getting a crash when I'm setting a vec3 now because this is expecting vec4.

stevk: Nevermind. I didn't realize that the w data would be thrown out when the enum is set to vec3.

kcoley: `MeshPrimitive` will take Vector4 as color, but you can specify in the accessor enums what kind of Vector you want in the final glTF file (it will ignore the `w` component if you want Vec3)

PR #131, opened 2017-09-20
Adding byte padding logic for color and texture coords

kcoley:
This should ensure the byte length should be a multiple of four when using normalized unsigned byte or unsigned short


kcoley:
I just realized I need to also offset the bytes when doing the byte padding.  I'll do a quick PR bug fix for this.

kcoley:
May not be as quick as I thought.. Running some more tests.

PR #133, opened 2017-09-21
fixed bug in incorrect byte size

kcoley:
I had a typo in calculating the bytelength for color in the last PR so this should fix it.

Will work on another PR to refactor the byte writing logic to make it more efficient.

### DIFF
--- a/Source/Runtime/MeshPrimitive.cs
+++ b/Source/Runtime/MeshPrimitive.cs
@@ -15,7 +15,7 @@ namespace AssetGenerator.Runtime
         /// Specifies which mode to use when defining the color accessor (Float is the default value)
         /// </summary>
         [Flags]
-        public enum ColorAccessorModeEnum { FLOAT = 0x000, NORMALIZED_USHORT = 0x001, NORMALIZED_UBYTE = 0x002, VEC3 = 0x100, VEC4 = 0x200 };
+        public enum ColorAccessorModeEnum { FLOAT = 0x001, NORMALIZED_USHORT = 0x002, NORMALIZED_UBYTE = 0x004, VEC3 = 0x100, VEC4 = 0x200 };
         /// <summary>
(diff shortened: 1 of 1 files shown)


Source/Runtime/MeshPrimitive.cs
```diff
@@ -15,7 +15,7 @@ public class MeshPrimitive
         /// Specifies which mode to use when defining the color accessor (Float is the default value)
         /// </summary>
         [Flags]
-        public enum ColorAccessorModeEnum { FLOAT = 0x000, NORMALIZED_USHORT = 0x001, NORMALIZED_UBYTE = 0x002, VEC3 = 0x100, VEC4 = 0x200 };
+        public enum ColorAccessorModeEnum { FLOAT = 0x001, NORMALIZED_USHORT = 0x002, NORMALIZED_UBYTE = 0x004, VEC3 = 0x100, VEC4 = 0x200 };
```
bghgary: This change doesn't make sense to me. You can't be a FLOAT and NORMALIZED_USHORT and NORMALIZED_UBYTE at the same time. What you had before makes more sense.

kcoley: I was trying to get this line to work when calculating the boolean for normalize since it seemed more readable:
`bool normalized = (ColorAccessorMode & ColorAccessorModeEnum.FLOAT) != ColorAccessorModeEnum.FLOAT;`

Though the original version does ensure that you cannot have invalid conditions.  I'll revert that and just use boolean operations to determine normalization.

bghgary: > you cannot have invalid conditions

Not exactly. Because of the `[Flags]` attribute, you can still put `FLOAT | NORMALIZED_USHORT`. The only way to really prevent invalid values is to list out all the possibilities or split the enum.

PR #146, opened 2017-09-29
Implement the new OcclusionRoughnessMetallic texture

stevk:


Source/TestValues.cs
```diff
@@ -15,7 +15,7 @@ public class TestValues
         private List<List<Parameter>> removeCombos = new List<List<Parameter>>();
         bool onlyBinaryParams = true;
         bool noPrerequisite = true;
-        string texture = "UVmap2017.png";
+        string texture = "TexturePlane_OcclusionRoughnessMetallic.png";
```
bghgary: You need to use a different texture for each texture property of the material except occlusion and metallicRoughness both point to the same texture.

stevk: Oh, right. This will be a slightly more extensive change, so I'd like to merge my refactor so I don't need to add the textures more than once.

Issue #134, opened 2017-09-21
Update Reference Textures


kcoley:
@patrickryanms has graciously set up some textures for glTF that he has allows us to use for this project:

https://github.com/PatrickRyanMS/TextureTests/tree/master/glTF

Should have all the passes we need for testing

Issue #98, opened 2017-09-14
Break Program.cs and TestValues.cs into more files and functions


stevk:
These changes should make the code more readable and easier to update.

- [x] The if statement trees in program.cs should each have their own function/file.
- [x] The helper functions in TestValues.cs should have their own file.
- [x] The combo filtering in TestValues.cs should be in its own function.

PR #149, opened 2017-09-29
Refactor TestValues.cs and Program.cs into more logical classes

stevk:
Really big change that will make additions and improvements easier, but doesn't change the output.
- Each test (Material, Material_Alpha, etc...). has its own class which contains its own values and if statements for building models. All of these are in the 'Tests' folder.
- Log file building has its own class, and the logic I used to make the table strings more human readable are in their own static class.
- The TestValue enums are now in their own file at the namespace level.
- Class for tracking the different parts of an attribute are in their own file now.
- Logic for building combos is in its own static class

In addition
- I've switched from using the word "Parameter" to "Attribute". (I had been using both inconsistently, and parameter seemed more confusing).
- I've replaced some abbreviations in an effort to make variables more readable (mat -> material).
- Some overloaded functions were replaced with optional parameters.


kcoley:
- [ ] Rename attribute.cs to something else. Maybe property
- [ ] Use switches instead of if else. Especially on the longer ones.
- [ ] Use reflection instead of if statements to run tests
- [ ] Move enums into attributes.cs and testvalues.cs
- [ ] Use reflection to set textures instead of explicitly setting them

stevk:
I only moved some of the tests from if/else statements to using switches, because they were evaluating multiple values in a single check.

Source/AssetGeneratorTests/Matrix4x4Tests.cs
```diff
@@ -6,7 +6,7 @@
 using System.Text;
 using System.Threading.Tasks;
 
-namespace AssetGenerator.Tests
+namespace AssetGenerator.TestNames
```
bghgary: Was this intentional?

Source/Attributes.cs
```diff
@@ -3,29 +3,18 @@
 namespace AssetGenerator
 {
     [AttributeUsage(AttributeTargets.Class)]
-    public class AssetGroupAttribute : Attribute
+    public class TestAttribute : System.Attribute
     {
-        public string Folder { get; private set; }
+        public TestNames TestName { get; private set; }
```
bghgary: TestNames enum can be singular. TestName property can just be Name.

Source/Tests/Material.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System.Collections.Generic;
+
+namespace AssetGenerator.Tests
+{
+    [TestAttribute(TestNames.Material),
```
bghgary: Move metadata into TestValues class. TestValues should probably be renamed to Test or TestBase.

stevk:
Feedback actioned.

Source/Attributes.cs
```diff
@@ -5,22 +5,22 @@ namespace AssetGenerator
     [AttributeUsage(AttributeTargets.Class)]
     public class TestAttribute : System.Attribute
     {
-        public TestNames TestName { get; private set; }
+        //public TestName Name { get; private set; }
```
bghgary: remove commented out code?

Source/Tests/Material.cs
```diff
@@ -2,15 +2,12 @@
 
 namespace AssetGenerator.Tests
 {
-    [TestAttribute(TestNames.Material),
-        ImageAttribute(texture_Emissive),
-        ImageAttribute(texture_Normal),
-        ImageAttribute(texture_OcclusionRoughnessMetallic)]
-    class Material : TestValues
+    [TestAttribute()]
```
bghgary: I think you can omit the `()`

Source/Attributes.cs
stevk: I could leave off the "Attribute" bit too, according to VS. Not sure if I should.

bghgary: Meh, I usually put it.

stevk:
Fixed the following issues
#134 
#98

PR #151, opened 2017-09-29
Added asset and extras to runtime abstraction

kcoley:
Added Asset to runtime abstraction

### DIFF
--- a/Source/Program.cs
+++ b/Source/Program.cs
@@ -146,7 +146,7 @@ namespace AssetGenerator
                         }
                     };
 
-                    var gltf = new Gltf
+                    var gltf = new glTFLoader.Schema.Gltf
                     {
                         Asset = asset.ConvertToAsset()
                     };
--- a/Source/Program.cs
+++ b/Source/Program.cs
@@ -9,10 +9,6 @@ using System.Linq;
 
 namespace AssetGenerator
 {
-    class ExtraData: Extras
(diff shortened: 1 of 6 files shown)


Source/Program.cs
```diff
@@ -140,24 +136,27 @@ private static void Main(string[] args)
...
+                    var gltf = new Gltf
+                    {
+                        Asset = asset.ConvertToAsset()
+                    };
+
                     var dataList = new List<Data>();
 
                     var geometryData = new Data(test.ToString() + "_" + comboIndex + ".bin");
                     dataList.Add(geometryData);
                     Runtime.GLTF wrapper = Common.SinglePlane(gltf, geometryData);
```
stevk: I'm not sure why this line compiled, but I'm pretty sure `gltf` needs to be replaced by `asset` here.

There also is a similar line near the end of the file `glTFLoader.Interface.SaveModel(gltf, assetFile);`

kcoley: Good catch, I'll fix that

kcoley: Turns out we are not even using gltf and geometryData in Common anymore.  I'll make a separate PR to remove those arguments from SimplePlane and other models.

kcoley: I rebased the update to remove `gltf` and `geometryData` from the Common args.

PR #169, opened 2017-10-05
Make changes with alpha attributes more apparent and remove unneeded emissive texture model

stevk:


stevk:
Also added a test for AlphaMask + Texture with Alpha + AlphaFactor

PR #170, opened 2017-10-05
Add models for v0.2

stevk:


kcoley:
Which models are being added for this PR?

stevk:
These were the models we reviewed during yesterday's meeting. So the new test model sets are:
Primitive_Attribute
Texture_Sampler

We've also split one of the previous sets in to new sets:
Material_Alpha
Material

And all of the test model sets have had individual models added/removed.

stevk:
These properties are not resulting in visible changes. 

Material
- Normal Scale
- Occlusion Strength

Material-Alpha
- Blend vs Mask
- Mask w/ factor

bghgary:
Closing. v0.2 has been committed directly.

Issue #174, opened 2017-10-05
Alpha cut off number gets clamped at 2 digets, cutting off value


stevk:

stevk:
Get a texture that a value with only one significant digit is needed to see a difference from .5

PR #178, opened 2017-10-05
Increace alpha cuttoff value enough so that we can see the significant figure in the md table

stevk:
MD table clamps at two digits, so dropped a 0 and moved it up to 0.2


stevk:
Fixes bug #174

PR #182, opened 2017-10-06
use four vertices instead of 6 for overrides

kcoley:
This adds in the update for the color bug fix and switches the primitive attribute tests to use four values instead of 6 for the plane


stevk:
Is this because we're using indices now?

kcoley:
Yeah, that's right.  With indices, we no longer need to duplicate the third and the sixth vertex

Issue #206, opened 2017-10-11
The way we're creating MD tables isn't flexible enough


stevk:
Add a template MD file for each test, where we stream that file into data and then insert a table. Each template will contain text specific to that test.

PR #209, opened 2017-10-11
Use templates for creating MD logs

stevk:
Gives each test a .md file which it uses as a template for creating its MD log file. As far as output, this PR should only be a refactor.

Doing this allows us to insert whatever text we want into each test's log, In the short term we should use this change to explain weird situations (Like models that look the same), and longer term we should be able to easily leverage this to insert extra images (e.g. to show UV mapping).

Note that a test determines its template by converting its testType into a string, so the name of a template must match that.

`~~HeaderTable~~` is the tag which is replaced when inserting the 'In all models' table.
`~~Table~~` is the tag which is replaced when inserting the main table.

Fixes issue #206


Source/AssetGenerator.csproj
```diff
@@ -81,6 +81,11 @@
   <ItemGroup>
     <None Include="App.config" />
     <None Include="packages.config" />
+    <EmbeddedResource Include="Tests\Material.txt" />
```
bghgary: Does renaming them to md not work? That would be a bummer if I can't use a markdown editor extension in VS to edit them.

stevk:
Switched them to .md

stevk:
Moved the images into the Tests folder and started using them as embedded resources. This should prevent broken link issues during editing if we add images into the templates.

PR #214, opened 2017-10-16
Add Material_SpecularGlossiness test set

stevk:
Adds 16 new models testing the SpecGloss extension.


Source/AssetGenerator.csproj
```diff
@@ -96,13 +96,13 @@
     <EmbeddedResource Include="Tests\lambert2_baseColor.png" />
     <EmbeddedResource Include="Tests\lambert2_emissive.png" />
     <EmbeddedResource Include="Tests\lambert2_normal.png" />
-    <EmbeddedResource Include="Tests\lambert2_occlusionRoughnessMetallic.png" />
+    <EmbeddedResource Include="Tests\lambert2_MetallicRoughness.png" />
```
bghgary: should be lower camel case to match other textures (i.e. MetallicRoughness -> metallicRoughness)

kcoley:
I think you will need to set the specularFactor to [0, 0, 0] when you generate test models that are not using specular.  Otherwise, the specularFactor defaults to [1,1,1] and overrides the diffuse texture

Source/Test.cs
```diff
@@ -16,7 +16,9 @@ internal class Test
         public const string texture_Normal = "lambert2_normal.png";
         public const string texture_Emissive = "lambert2_emissive.png";
         public const string texture_BaseColor = "lambert2_baseColor.png";
-        public const string texture_OcclusionRoughnessMetallic = "lambert2_occlusionRoughnessMetallic.png";
+        public const string texture_MetallicRoughness = "lambert2_metallicRoughness.png";
```
bghgary: Maybe we should separate out occlusion texture into its own texture.

stevk:
Is this what we had in mind for occlusion?

image

stevk:
image

stevk:
- [ ] Only test float-vec3 vertex color in metallicRough/specgloss
- [ ] Test 4 vertex color in alpha
- [ ] Large pinapple

stevk:
MetallicRoughness
image

Spec Gloss
image

Alpha
image

PR #230, opened 2017-10-27
Convert Plane model to +Z forward

kcoley:


bghgary:
See https://github.com/KhronosGroup/glTF/issues/1043#issuecomment-339132697

PR #232, opened 2017-10-31
Compatibility Model Test set

stevk:
Adding six models to test glTF compatibility.

Model_00 is a control, as usual.

Models 01-04 tests setting the MinVersion and Version fields, as well as having code in the JSON that glTF 2.0 won't recognize.

Model_05 tests having an required extension which isn't supported by the client.

image


stevk:
image

Source/Runtime/Extensions/MicrosoftQuantumRendering.cs
```diff
@@ -0,0 +1,207 @@
+﻿// This is a fake extension
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace AssetGenerator.Runtime.Extensions
+{
+    internal class MicrosoftQuantumRendering : Extension
```
kcoley: Maybe not call this `MicrosoftQuantumRendering` and just call it `QuantomRendering`, so not to associate a fake extension with a company.

Source/Runtime/Extensions/MicrosoftQuantumRendering.cs
```diff
@@ -0,0 +1,207 @@
+﻿// This is a fake extension
```
kcoley: This header could explain why we are introducing a fake extension in the asset generator code-base and to ensure no one actually uses it for anything.

Source/Schema/MaterialMicrosoftQuantumRendering.cs
```diff
@@ -0,0 +1,267 @@
+﻿namespace glTFLoader.Schema
+{
+    using System.Linq;
+    using System.Runtime.Serialization;
+
+
+    public class MaterialMicrosoftQuantumRendering
```
kcoley: Same comment about not using a company name for a fake material: `MaterialQuantumRendering`, instead of `MaterialMicrosoftQuantumRendering`.

Source/Runtime/Extensions/MicrosoftQuantumRendering.cs
stevk: @bghgary input? It was at your suggestion that I added Microsoft to the name of the fake extension.

Source/Runtime/Extensions/MicrosoftQuantumRendering.cs
stevk: @bghgary input? It was at your suggestion that I added Microsoft to the name of the fake extension.

Source/Runtime/Extensions/MicrosoftQuantumRendering.cs
stevk: Good idea. Updated.

stevk:
- [ ] Add 'In 2.0' to 'Model Should Load'
- [ ] Switch enum from Sampler to AlphaMode 'Multiply' by adding a new alphamode field
- [ ] Replace "Microsoft" with "Ext_"
- [ ] Remove caps from root lights property
- [ ] Remove the texture from the extensions required test. Quantum rendering is just that good.
- [ ] Be verbose on what the extra features are doing in the table
- [ ] Rename 'Experimental Features' to something else, and combine it with 'Extension Required'
- [ ] Don't show the fallback sampler and texture. Just explain.

bghgary:
> Replace "Microsoft" with "Ext_"

Must be all caps for the prefix: EXT_

stevk:
Good catch, thanks!

stevk:
I'm pushing back on the `Add 'In 2.0' to 'Model Should Load'` feedback. Adding a period would break the enum, so I'd need to dramatically change how the code is generating table headers to make this work. I'm also not convinced that it adds valuable information, especially now that we're including the Extension Required property in the same column.

bghgary:
> pushing back on the Add 'In 2.0' to 'Model Should Load' feedback

The problem for me is that a 2.1 loader should be able to load the 2.1 asset, so saying the model should not load for that asset is misleading once we have 2.1 loaders. Maybe incorporate that column into the description column?

stevk:
https://gist.github.com/stevk/36b1f6f91109755cc418be05dee16b97

image

Source/Runtime/Extensions/EXT_QuantumRendering.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// Converts the material to schema
+        /// </summary>
+        /// <param name="gltf"></param>
+        /// <param name="samplers"></param>
+        /// <param name="images"></param>
+        /// <param name="textures"></param>
+        /// <returns></returns>
+        public override Object ConvertToSchema(Runtime.GLTF gltf, List<glTFLoader.Schema.Sampler> samplers, List<glTFLoader.Schema.Image> images, List<glTFLoader.Schema.Texture> textures)
+        {
+            glTFLoader.Schema.MaterialEXT_QuantumRendering materialMicrosoftQuantumRendering = new glTFLoader.Schema.MaterialEXT_QuantumRendering();
```
kcoley: Can we change this to not use Microsoft as well?

stevk:
Good catch, I thought I had fixed that everywhere.

Source/Runtime/Extensions/EXT_QuantumRendering.cs
stevk: Good catch, I thought I had fixed that everywhere.

stevk:
https://gist.github.com/stevk/36b1f6f91109755cc418be05dee16b97

image

bghgary:
* Change "Simulated Feature" to "Description"
* Make Description column lower cased
* Remove X from model 04
* Remove top two lines warning
* "ignore new enum" should be "ignore new property"
* extra comma after "and, "
* "max version" can be rephrased as "highest supported version"

stevk:
https://gist.github.com/stevk/36b1f6f91109755cc418be05dee16b97

image

PR #236, opened 2017-11-01
update for v0.3.1

kcoley:
Changed:

- corrected normalized byte and short values in primitive attribute tests
- restored note about normalized byte and short in the primitive attribute readme
- changed models to +z forward convention


bghgary:
Is 255 commits correct? Seems a bit large.

stevk:
I think it's listing all of the commits since we created develop, but the delta is really only the last ~6 commits. Maybe a consequence of not deleting develop after each merge to master?

bghgary:
Let's go over it offline.

PR #245, opened 2017-11-08
Add enum for defining different index component types

kcoley:
Add support for defining different indice types using enum.

### DIFF
--- a/Source/Runtime/MeshPrimitive.cs
+++ b/Source/Runtime/MeshPrimitive.cs
@@ -61,6 +61,16 @@ namespace AssetGenerator.Runtime
         public List<Vector4> Tangents { get; set; }
 
         /// <summary>
+        /// Available component types to use when defining the indices accessor
+        /// </summary>
+        public enum IndexComponentTypeEnum { UNSIGNED_INT, UNSIGNED_BYTE, UNSIGNED_SHORT };
+
+        /// <summary>
+        /// Specifices which component type to use when defining the indices accessor
+        /// </summary>
+        public IndexComponentTypeEnum IndexComponentType { get; set; }
+
+        /// <summary>
         /// List of indices for the mesh primitive
         /// </summary>
         public List<int> Indices { get; set; }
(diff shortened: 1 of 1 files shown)


Source/Runtime/MeshPrimitive.cs
```diff
@@ -61,6 +61,16 @@ public enum TextureCoordsComponentTypeEnum { FLOAT, NORMALIZED_USHORT, NORMALIZE
         public List<Vector4> Tangents { get; set; }
 
         /// <summary>
+        /// Available component types to use when defining the indices accessor
+        /// </summary>
+        public enum IndiceComponentTypeEnum { UNSIGNED_INT, UNSIGNED_BYTE, UNSIGNED_SHORT };
```
bghgary: Indice --> Index?

Source/Runtime/MeshPrimitive.cs
```diff
@@ -280,7 +290,21 @@ public glTFLoader.Schema.MeshPrimitive ConvertToSchema(Runtime.GLTF gltf, List<g
...
+
+                switch (IndiceComponentType)
+                {
+                    case IndiceComponentTypeEnum.UNSIGNED_BYTE:
+                        indiceComponentType = glTFLoader.Schema.Accessor.ComponentTypeEnum.UNSIGNED_BYTE;
+                        break;
+                    case IndiceComponentTypeEnum.UNSIGNED_SHORT:
+                        indiceComponentType = glTFLoader.Schema.Accessor.ComponentTypeEnum.UNSIGNED_SHORT;
+                        break;
+                    default: //Defaults to glTFLoader.Schema.Accessor.ComponentTypeEnum.UNSIGNED_INT;
```
bghgary: nit: I would prefer switch statements to include all the enum values and have a default that throws. That way if a new enum value is added, this code will fail explicitly.

kcoley: gotcha

PR #248, opened 2017-11-08
removing unused spec gloss classes

kcoley:


kcoley:
Removes these two unused classes from the project.  They are already not referenced in the Visual Studio solution.

PR #250, opened 2017-11-08
Adding Node class to runtime abstraction

kcoley:


Source/Runtime/Node.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        /// <summary>
+        /// The mesh in this node.
+        /// </summary>
+        public Mesh Mesh { get; set; }
+
+        /// <summary>
+        /// The node's unit quaternion rotation in the order (x, y, z, w), where w is the scalar.
+        /// </summary>
+        public Quaternion Rotation { get; set; }
```
bghgary: How come this is not nullable?

Source/Runtime/Node.cs
```diff
@@ -0,0 +1,108 @@
...
+{
+    /// <summary>
+    /// A node in a node hierarchy.  
+    /// </summary>
+    internal class Node
+    {
+        /// <summary>
+        /// A floating-point 4x4 transformation matrix stored in column-major order.
+        /// </summary>
+        public Matrix4x4 Matrix { get; set; }
```
bghgary: This should also be nullable

Source/Runtime/GLTF.cs
```diff
@@ -63,45 +63,18 @@ public void BuildGLTF(ref glTFLoader.Schema.Gltf gltf, Data geometryData)
...
+            int bufferIndex = 0;
 
 
             // for each scene, create a node for each mesh and compute the indices for the scene object
             foreach (Runtime.Scene gscene in Scenes)
             {
                 List<int> scene_indices_set = new List<int>();
                 // loops through each mesh and converts it into a Node, with optional transformation info if available
-                for (int mesh_index = 0; mesh_index < gscene.Meshes.Count(); ++mesh_index)
+                for(int node_index = 0; node_index < gscene.Nodes.Count(); ++node_index)
```
bghgary: node_index -> nodeIndex

Source/Runtime/Node.cs
kcoley: Currently `Quaternion` is implemented as a class.  When we switch to using the C# Numerics, we can convert these all to nullable.

Source/Runtime/Node.cs
kcoley: Currently `Matrix4x4` is implemented as a class.  When we switch to using the C# Numerics, we can convert these all to nullable.

Source/Runtime/GLTF.cs
kcoley: Missed this one

PR #253, opened 2017-11-09
Implement Mesh test set

stevk:
Adding 34 models in two Mesh test sets. Mesh_Primitives and Mesh_Indices.

Mesh_Primitives tests having multiple primitives in one mesh, and applying primitive attributes to the primitives in varying degrees.

Mesh_Indices
 Tests primitive mode, indices component types, and using/not using indices.

Outstanding issues:
- Mesh_Primitives MD log has a lot of columns, and probably won't fit nicely to a screen.

**Mesh_Primitives** : https://gist.github.com/stevk/65f6f52b6e5634b519c0d3bc5e735a9d

image

**Mesh_Indices** : https://gist.github.com/stevk/56c1f1aa4bb63125ee7a4a7a138dfa8b

image


stevk:
- [ ] Remove 'Normalized' from the table header. They're just unsigned.
- [ ] Split into two tests, Mesh_Indices and Mesh_Primitives
- [ ] Make combos for every primitive mode with and without indices

stevk:
I've updated the image of the log. I'm still working on verifying that the attributes are being set to the models correctly.

stevk:
Added the nice icons to the logs and added a little more text. Ready to review.

Source/Tests/Mesh_Indices.md
```diff
@@ -0,0 +1,17 @@
+No attributes are explicitly set in every model generated by this test, other than those in the base model.  
+
+All values of Byte, SHort, and Int are unsigned
```
kcoley: `SHort` should be `Short`

Source/Tests/Mesh_Indices.md
```diff
@@ -0,0 +1,17 @@
...
+
+<br>
+
+All model's indices values relate to vertexes as shown by the following image.
```
kcoley: Should be `All model indices relate to vertices as shown by the following image:`

Source/Tests/Mesh_Primitives.md
```diff
@@ -0,0 +1,23 @@
+No attributes are explicitly set in every model generated by this test, other than those in the base model.  
+
+<br>
+
+Both primitives are using the same indices values, but have different vertex positions.
```
kcoley: Should be `Both primitives are using the same index values, but have different vertex positions.`

Source/Tests/Mesh_Primitives.md
```diff
@@ -0,0 +1,23 @@
...
+
+Both primitives are using the same indices values, but have different vertex positions.  
+
+Indices for Primitive 1 (Left) | Indices for Primitive 2 (Right)
+:---: | :---:
+<img src="./Icon_Indices_Primitive1.png" height="72" width="72" align="middle"> | <img src="./Icon_Indices_Primitive2.png" height="72" width="72" align="middle">
+
+<br><br>
+
+If Vertex UV 1 is set, then it is the UV used by the texture. Otherwise Vertex UV 0 is used. Both primitives have different UV values.
```
kcoley: nit: Maybe comma after Otherwise

kcoley:
For Mesh_Primitives test, change text under the primitives column to be a little briefer:
 - Neither has attributes set -> No attributes set
 - Both have attributes set -> Both attributes set
 - First (on left) has attributes set -> Left primitive attributes set
 - Second (on right) has attributes set -> Right primitive attributes set

Two primitives may be unnecessary since it is consistent across all entries

kcoley:
Why does the Mesh_Primitive README.md file have two columns for Vertex UV 0 with the same row values?

sbtron:
LGTM

Source/Tests/Mesh_Indices.md
stevk: Fix pending

Source/Tests/Mesh_Indices.md
stevk: Fix pending

Source/Tests/Mesh_Primitives.md
stevk: Fix pending

Source/Tests/Mesh_Primitives.md
stevk: Fix pending

stevk:
@kcoley The identical columns are for each of the primitives having their own UV mapping. I can add a better column header to label this, but this also brings up another point. There currently isn't a model where the primitives are using different UVs, both are either 0 or 1 (even if the values of those are different). I think I'll add another two models to cover that.

kcoley:
- [ ] Remove texture coordinates from mesh indices

stevk:
@kcoley I think I found a bug. I'm, setting the TexCoordIndex for one primitive, but it is being changed for both. Line 421 of Mesh_Primitives. I'll come over to try and debug it in person as soon as I can.

PR #257, opened 2017-11-10
Refactoring schema conversion logic to GLTFConverter class.

kcoley:
- Refactors schema conversion logic to `GLTFConverter` class.
- Replaces `BuildGLTF` method with static method `Runtime.GLTFConverter.ConvertRuntimeToSchema`
- Sets the indices of the children of the `Node` class at the top level (parent node has the lower index).


kcoley:
This PR also sets the children of the node at the top level (parent node has the lower index).

Source/Program.cs
```diff
@@ -78,7 +78,8 @@ private static void Main(string[] args)
 
                     // Passes the desired properties to the runtime layer, which then coverts that data into
                     // a gltf loader object, ready to create the model
-                    wrapper.BuildGLTF(ref gltf, geometryData);
+                    Runtime.GLTFConverter.ConvertRuntimeToSchema(wrapper, ref gltf, geometryData);
+                    //wrapper.BuildGLTF(ref gltf, geometryData);
```
stevk: Remove commented out code.

Source/Runtime/GLTFConverter.cs
```diff
@@ -0,0 +1,1252 @@
...
+            if (specGloss.GlossinessFactor.HasValue)
+            {
+                materialPbrSpecularGlossiness.GlossinessFactor = specGloss.GlossinessFactor.Value;
+            }
+
+            return materialPbrSpecularGlossiness;
+        }
+
+        /// <summary>
+        /// Converts Runtime Quatum Rendering to Schema (Not an actual glTF feature)
```
stevk: Change 'Quatum' to 'Quantum'

Source/Program.cs
kcoley: good catch

Source/Runtime/GLTFConverter.cs
kcoley: good catch

PR #263, opened 2017-11-16
Fix for making matrix nullable in Runtime.Node

kcoley:


kcoley:
Super small fix, so merging directly.

PR #269, opened 2017-11-18
Adding Metallic Roughness Fallback to Specular Glossiness Tests

kcoley:


Source/Tests/Material_SpecularGlossiness.cs
```diff
@@ -6,6 +6,7 @@ namespace AssetGenerator.Tests
     [TestAttribute]
     class Material_SpecularGlossiness : Test
     {
+        private Runtime.Image MetallicRoughnessFallbackTexture { get; set; }
```
stevk: This should instead be added as a property in the `specialProperties` list, and then inserted into each combo via the `ApplySpecialProperties` method to be consistent.

Source/Tests/Material_SpecularGlossiness.cs
```diff
@@ -165,8 +171,21 @@ public Runtime.GLTF SetModelAttributes(Runtime.GLTF wrapper, Runtime.Material ma
...
+            // Fallback PBR Metallic Roughness
+            var pbrMetallicRoughnessFallback = new Runtime.PbrMetallicRoughness
+            {
+                BaseColorTexture = new Runtime.Texture
+                {
+                    Source = MetallicRoughnessFallbackTexture,
+                    Name = "Metallic Roughness Fallback Texture"
+                }
+            };
+            material.MetallicRoughnessMaterial = pbrMetallicRoughnessFallback;
```
stevk: This works, but should instead be added as a case after the original value is made into a special property.

Source/Tests/Material_SpecularGlossiness.cs
```diff
@@ -165,8 +171,21 @@ public Runtime.GLTF SetModelAttributes(Runtime.GLTF wrapper, Runtime.Material ma
...
             wrapper.Scenes[0].Nodes[0].Mesh.MeshPrimitives[0].Material = material;
 
+            
+
```
stevk: Remove the extra lines and spaces that were added here.

Source/Tests/Material_SpecularGlossiness.md
```diff
@@ -1,4 +1,6 @@
 No attributes are explicitly set in every model generated by this test, other than those in the base model.  
+
+If the engine does not support the Specular Glossiness extension, the models will fall back to Metallic Roughness and render an "X" texture as the base color texture.
```
stevk: The first line should be changed to match this, since now we are explicitly setting this attribute in every model.

kcoley:
@stevk , thanks for the feedback.  I'll work on those changes.

kcoley:
@stevk , updates rolled out:
 - Readme.md file has been updated with displaying the base color texture
 - Now using requiredProperties for base color texture
 - Some minor code changes in `Material_SpecularGlossiness` (switching to var and removing extra semicolons in switch statement)

PR #275, opened 2017-11-22
Stevk/specglossupdate

### LINKED ISSUE #268: Add an asset with two materials - one with specgloss and metalrough and another just metalrough
Ran into this with UnityglTF loader not being able to handle this use case.
Should be a valid asset with specgloss extensionused but not necessary that all materials in the asset also define the specgloss material.

stevk:
Added two new models to spec gloss. One where spec gloss was set on only one of two meshes, and another where spec gloss wasn't set except in the `Extension Used` field.

 Also set `Specular Glossiness Extension Used` on all of the models.

As a part of this work, I replaced the deep copy function with another. The original was having issues with System.Numerics when copying Vec3.

Fixes #268 

image


stevk:
Added the fixes we talked about. The double mesh model also wasn't working correctly, as it was only listing a single material.

PR #281, opened 2017-12-05
Mesh model group updates

stevk:
Mesh_Primitives
- Split UVs into another table with attributes on by default. 
- On the UV icons, show the UV coords on the corners
- Add models to cover using UV1 when there is no UV0

Mesh_Indices
- Show vertices values and communicate that we're using different vertices with vs without indices
- Group the "With Indices" together instead of alternating
- Hack in three models that have UV1 without UV0

Mesh_Indices
image

Mesh_Primitives
image

Mesh_PrimitivesUV
image


kcoley:
Mesh_PrimitivesUV_02 has incorrect `texCoord` index for material.  Should be 0, not 1.

stevk:
Good feedback Kcoley. I've made those changes.

Issue #279, opened 2017-11-29
Material_Alpha: Add double sided test models with normals


kcoley:
We have double sided tests models without normals, but we will also need models with normals.

Issue #278, opened 2017-11-28
update material alpha model 01 so that it is not completely transparent


kcoley:

stevk:
Fixed. If this is still transparent, then it is a bug with that client.

Issue #190, opened 2017-10-09
Material_Alpha - Use PBR and normal texture to make it look more like the basic texture


stevk:
future?

stevk:
A base color texture was added where it made sense. Do we still want to add a normal texture?

stevk:
We're only going to use textures on a few of these models.

PR #295, opened 2017-12-18
Add models to Material_Alpha. A range of Alpha cutoff values, and doublesided with normals

### LINKED ISSUE #294: Test alpha Cutoff against more values
We should test the following combos:
- [x] Mask + Alpha cutoff + base color factor (same alpha value)
- [x] Mask + Alpha cutoff + base color texture (same alpha value)
- [x] Mask + Alpha cutoff + base color texture + base color factor (same alpha value)
- [x] Mask + Alpha cutoff + base color factor (Cutoff value is lower)
- [x] Mask + Alpha cutoff + base color texture (Cutoff value is lower)
- [x] Mask + Alpha cutoff + base color texture + base color factor (Cutoff value is lower)
- [x] Mask + Alpha cutoff + base color factor (Cutoff value is higher)
- [x] Mask + Alpha cutoff + base color texture (Cutoff value is higher)
- [x] Mask + Alpha cutoff + base color texture + base color factor (Cutoff value is higher)

stevk:
Fixed
#294 _Test alpha Cutoff against more values_
#279 _Add double sided test models with normal_

I also looked into bug #278 it appears to me that the model is transparent in Babylon.js due to a bug with that client. The same model is visible in three.js 

This also starts to address #190 _Use PBR and normal texture to make it look more like the basic texture_

11 new models are added, and normals are added to 01. This changes the number used by most of the pre-existing models as well.
**New Models : 04-08, 14-16, 22-24**

[Material_Alpha](https://github.com/stevk/glTF-Asset-Generator/blob/stevk/alphaupdate/Output/Material_Alpha/README.md)

[Material_Doublesided](https://github.com/stevk/glTF-Asset-Generator/blob/stevk/alphaupdate/Output/Material_Doublesided/README.md)


stevk:
https://github.com/stevk/glTF-Asset-Generator/blob/stevk/alphaupdate/Output/Material_Alpha/README.md

bghgary:
We discussed offline. Let's put the double sided, normal, tangent stuff into its own table.

stevk:
Sort order:
Alpha mode > Number of properties > Order of columns

Issue #187, opened 2017-10-09
Make MD Cells a consistent height


stevk:

stevk:
As mentioned in the PR, this isn't possible. 

The next best solution is to include an item of the same height in every row, like an image of the expected model appearance. This will likely be solved at the same time as issue #301

PR #297, opened 2017-12-18
Set cell heights in the readme so that models with and without texture use the same amount of space

stevk:
Fix for issue #187 _Make MD Cells all the same height, now that textures are bigger_

Both the main table of models as well as the list of required properties will have a fixed cell height which matches the image height.

Below is a sample, but every readme will be affected by this change.
image


stevk:
This PR will need to be updated again once the other two pull requests go in, or visa-versa, since the other PRs also make changes to their readme files.

https://github.com/stevk/glTF-Asset-Generator/tree/stevk/cellheight/Output

stevk:
Also, it looks like this cell height fix doesn't work on github's markdown. I'll try something else

bghgary:
If this isn't easy to do with GitHub markdown, just don't worry about it.

stevk:
Looks like it might not be possible to do this way.
https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/sanitization_filter.rb

As an idea, by including an image of each model (Expected rendering?) we'd force the cells into a uniform size. Closing this PR in the meantime.

Issue #302, opened 2017-12-20
Split model groups into smaller logical groups


stevk:
Try to keep each model group into ~10 models.
Make sure that the most complex table still has an all-up model

Primitive_Attribute : Vertex Color, UV0, UV0+UV1

Issue #284, opened 2017-12-11
Add a statement of purpose to each model group


stevk:
Each log should be clear about what is being tested. Every log should have at least a sentence at the top which explains that model group's purpose.

- [x] Compatibility
- [x] Material
- [x] Material_AlphaBlend
- [x] Material_AlphaMask
- [x] Material_Doublesided
- [x] Material_MetallicRoughness
- [x] Material_SpecularGlossiness
- [x] Material_Mixed
- [x] Mesh_Indices
- [x] Mesh_Primitives
- [x] Mesh_PrimitivesUV
- [x] Primitive_Attribute
- [x] Texture_Sampler

PR #305, opened 2017-12-21
Primitive_Attribute : Split Vertex Color into a new model group and remove UV1 tests

### LINKED ISSUE #163: Primitive_Attribute has several properties that need multiple values set
Instead of passing one value and then setting the rest in the if statement, it would be better to make an object to hold all of the needed values. 

Probably contain the var type, vec type, and the vec coord. Should be flexabile enough to use for color or tex cords.

- [x] Primitive_VertexColor
- [ ] Material_AlphaBlend
- [ ] Material_MetallicRoughness
- [ ] Material_SpecularGlossiness
- [ ] Mesh_Primitives
- [ ] Mesh_PrimitivesUV

stevk:
- Trim Primitive_Attribute down to 9 models.
[Primitive_Attribute Readme](https://github.com/stevk/glTF-Asset-Generator/blob/stevk/vertexcolorsplit/Output/Primitive_Attribute/README.md)

- Create Primitive_VertexColor by pulling 6 models out of Primitive_Attribute
[Primitive_VertexColor Readme](https://github.com/stevk/glTF-Asset-Generator/blob/stevk/vertexcolorsplit/Output/Primitive_VertexColor/README.md)

Set up the readme for both model groups to be sorted by increasing complexity, and removed the unnecessary references to the base model.

Progress on #302 _Split model groups into smaller logical groups_ 

Also mostly fixed #163 _Primitive_Attribute has several properties that need multiple values set_
I'll need to do more fixes later to address uses of ColorVertex in other groups


stevk:
Add a note about the alpha value used in the readme

stevk:
Model 04 doesn't have a material?

PR #307, opened 2017-12-22
Split Material_SpecularGlossinessApplied out of Material_SpecularGlossiness

stevk:
Split some edge case tests out of the spec gloss model group. Also applied sorting by complexity to each of them and updated their readme headers.

[Material_SpecularGlossinessApplied](https://github.com/stevk/glTF-Asset-Generator/blob/stevk/specglosssplit/Output/Material_SpecularGlossinessApplied/README.md)

[Material_SpecularGlossiness](https://github.com/stevk/glTF-Asset-Generator/blob/stevk/specglosssplit/Output/Material_SpecularGlossiness/README.md)

More progress on #302 _Split model groups into smaller logical groups_ 


stevk:
Material_SpecularGlossinessApplied
- [ ] Change name to Material_Mixed
- [ ] Make header table text one line
- [ ] Split into twoo columns, Material 0 and Material 1
- [ ] Use two triangles instead of two planes

Spec Gloss
- [ ] Diffuse factor w/ Diffuse texture
- [ ] Gloss factor w/ Diffuse texture
- [ ] Put Diffuse texture colum by Diffuse Factor

PR #308, opened 2017-12-22
Material_MetallicRoughness : Apply sorting

stevk:
Sorted in order of complexity.

[Material_MetallicRoughness](https://github.com/stevk/glTF-Asset-Generator/blob/stevk/metroughsort/Output/Material_MetallicRoughness/README.md)

More progress on #302 _Split model groups into smaller logical groups_ 


stevk:
- [ ] Move basecolortexture to be by basecolorfactor

stevk:
@bghgary  Updated. I also put the texture further forward so that the factor type properties would stay grouped.

Issue #158, opened 2017-10-04
Convert project to .Net Standard


v-blipsc:
Convert project to be a .Net Standard 2.0 assembly to give a consistent multiple platform support.

.Net Standard 2.0 has been released to the public

stevk:
I'm not sure how to go about doing this. We should talk about this.

PR #315, opened 2018-01-05
Update App to .Net Core 2.0

stevk:
Switch the project from .Net Framework to .Net Core, in order to improve cross platform compatibility.
  
Fix for #158


kcoley:
Just noticed that there is an AssetGeneratorTests folder that should be deleted.

stevk:
What is the path? I think I might have already gotten it.

kcoley:
nvm, think it was only on my local visual studio for some reason

PR #322, opened 2018-01-12
Node model groups

stevk:
Model groups for performing transforms on nodes, and negative scale. 

[Node_Attribute](https://github.com/stevk/glTF-Asset-Generator/blob/stevk/node/Output/Node_Attribute/README.md)

[Node_NegativeScale](https://github.com/stevk/glTF-Asset-Generator/blob/stevk/node/Output/Node_NegativeScale/README.md)


stevk:
I've added Node_Material, which should address testing the use of a mesh and a material without vertex colors in the same model as one that does use vertex colors.

stevk:
For context, this is what the raw matrix data is looking like

image

kcoley:
- The links for Node_Attribute and Node_Material do not seem to be working
 - Is it possible to display the matrices without the brackets and M: ?

stevk:
I've update the links in this PR. I'll look into reformatting matrices.

PR #324, opened 2018-01-18
Fixed bug in padding calculation for interleaved buffers

kcoley:
- Fixes padding issues with interleaved buffers
- Also ensures normalized values are flagged as normalized in interleaved buffers

### DIFF
--- a/Source/Runtime/GLTFConverter.cs
+++ b/Source/Runtime/GLTFConverter.cs
@@ -315,15 +315,22 @@ namespace AssetGenerator.Runtime
         }
 
         /// <summary>
-        /// Pads the value of the values to ensure it is a multiple of size
+        /// Pads a value to ensure it is a multiple of size
         /// </summary>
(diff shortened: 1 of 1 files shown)


Source/Runtime/GLTFConverter.cs
```diff
@@ -935,6 +935,7 @@ private static glTFLoader.Schema.MaterialEXT_QuantumRendering ConvertExtQuantumR
                         Accessors.Add(positionAccessor);
                         attributes.Add("POSITION", Accessors.Count() - 1);
                     }
+                    totalByteLength = ZeroPadGeometryData(geometryData, totalByteLength);
```
bghgary: I would recommend doing something more generic like this:
https://github.com/bghgary/glTF-Shell-Extensions/blob/master/Source/glTF/StreamExtensions.cs

kcoley:
@bghgary I updated the `Align` code to be more generic.

PR #329, opened 2018-01-30
Add three.js preview.

donmccurdy:
Manifest looks good! 🙂

Haven't spent any time looking into the errors or tweaking lights yet, but here's a bit of code to render everything in three.js.

Demo: https://cdn.rawgit.com/donmccurdy/glTF-Asset-Generator/feat-threejs-preview/Preview/three.js/index.html

screen shot 2018-01-30 at 3 51 21 pm

### DIFF
--- a/Preview/three.js/handlebars-v4.0.11.js
+++ b/Preview/three.js/handlebars-v4.0.11.js
@@ -1,0 +1,4840 @@
+/**!
+
+ @license
+ handlebars v4.0.11
+
+Copyright (C) 2011-2017 by Yehuda Katz
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


bghgary:
I'm planning on moving the preview folder to https://github.com/bghgary/glTF-Assets-Viewer. The idea is that we will add some links to each of the README of each folder and it will show the results in various engines.

bghgary:
Though, I wasn't aware of https://cdn.rawgit.com. That might change things.

donmccurdy:
Cool! Yeah up to you what if anything to do with this. I'm looking forward to debugging the handful that aren't showing up, great set of test cases here. I'll be traveling for the next ~week so won't be getting to this more until after that.

bghgary:
@donmccurdy We've talked about how each engine will do this kind of testing in their own repo. Can we close this?

donmccurdy:
Sounds good 👍

Issue #331, opened 2018-02-01
Readme has broken links and needs to be updated


bghgary:

stevk:
Which links are broken?

bghgary:
Material Alphas at least. We need update that list whenever the output folder structure changes.

stevk:
Oh! I see what you mean. It is the main GitHub readme that's broken. I thought you meant the readme for the individual model groups. I'll get a fix in.

https://github.com/bghgary/glTF-Asset-Generator/blob/master/README.md

PR #332, opened 2018-02-01
Update the main readme with the new model groups

stevk:
Add links to all of the new model groups we've created. 

Also update the title for the feature roadmap section.

https://github.com/stevk/glTF-Asset-Generator/blob/mainreadme/README.md

Fix for #331


stevk:
Wait on this one. I'm going to add some code so that it automatically updates as model groups are added.

stevk:
Alright, it should be ready now. Keep in mind that the main readme is now in the ReadmeTemplates and is written out each time the build is run, so changes to that readme should be made to the version in the templates folder and NOT to the user facing file.

Issue #354, opened 2018-02-28
Readme value formatting for Node_Attribute doesn't match how it would look in JSON format


stevk:
The formatting of these need some tweaking.

•	Numbers should not have an “f” postfix.
•	Quaternions don’t need X, Y, Z, W prefixes and should use [] and not {}.

In general, the values should look exactly like how it looks in JSON form, except for matrices since matrices are hard to read in JSON form.

PR #356, opened 2018-03-01
Change the readme values to be more in line with how it looks in JSON format

stevk:
Removed the postfix 'f' character, and added custom formatting for quaternion values.

This change affects several readme files and not just Node_Attribute, due to all vectors using the postfix 'f'.

Fix for #354


Source/ReadmeStringHelper.cs
```diff
@@ -91,6 +91,16 @@ public static string ConvertTestValueToString(Property param)
...
                 }
+                else if (valueType.Equals(typeof(Quaternion)))
+                {
+                    string[] quaternionValues = new string[4];
+                    quaternionValues[0] = param.value.X.ToString("N1");
+                    quaternionValues[1] = param.value.Y.ToString("N1");
+                    quaternionValues[2] = param.value.Z.ToString("N1");
+                    quaternionValues[3] = param.value.W.ToString("N1");
+
+                    output = String.Format("[{0}, {1}, {2}, {3}]", quaternionValues).Replace(" ", "&nbsp;");
```
bghgary: this can be simplified to `string.Format("[{0:N1}, {1:N1}, {2:N1}, {3:N1}", quaternionValues)` and change quaternionValues to float[4];

bghgary: You can also inline the replace(" ", "&nbsp;")

stevk: I'm not sure I understand what you mean by 
> inline the replace(" ", " ")
Isn't `.Replace(" ", "&nbsp;"); ` already inline?

bghgary: I meant the latter. It's fine to not do it.

stevk: The change has been made. It isn't quite as simple as we could have hoped for due to C# not supporting co-variant array conversion from int[] to object[]

Issue #357, opened 2018-03-05
The node attribute transforms are not framed within the screenshot camera


kcoley:
Use different set of node transforms.

kcoley:
default:
translation: 0, 0, -1.3
rotation: 0, -1, 0, 0

node negative scale: 
translation: 0, 20, 20
rotation: -0.4, 0, 0, 1

node attribute 
I used negative scale one

PR #360, opened 2018-03-05
Change Node_Attribute and camera values to facillitate creating sample images

stevk:
- Updates the camera values written to the manifest.
- Textures the Node_Attribute models, to make it easier to see what is going on.
- Changes the scale and translate values for Node_Attribute, so the model states on-camera and there is no z-fighting or clipping.
- Removes the reference node. We can add parent-child hierarchy testing later.

This is a fix for #357 
Contains some fixes for #363


kcoley:
- [ ] Moved matrix column to the right and move matrix test model to the bottom
- [ ] Use a positive y-translation value instead of -4.1
- [ ] Ensure the matrix transform is identical to the translation/rotation and scale

Source/ModelGroups/Node_Attribute.cs
```diff
@@ -12,22 +12,32 @@ public Node_Attribute(List<string> textures, List<string> figures) : base(textur
...
+            {
+                Uri = textures.Find(e => e.Contains("BaseColor_Nodes"))
+            };
+            usedTextures.Add(baseColorTexture);
+
+            var matrixT = Matrix4x4.CreateTranslation(new Vector3(2, 2, 2));
             var matrixR = Matrix4x4.CreateFromAxisAngle(new Vector3(0f, 1f, 0f), (float)(Math.PI));
-            var matrixS = Matrix4x4.CreateScale(2);
+            var matrixS = Matrix4x4.CreateScale(1.2f);
             var matrixTRS = Matrix4x4.Multiply(Matrix4x4.Multiply(matrixT, matrixR), matrixS);
```
kcoley: S * R * T

PR #362, opened 2018-03-06
Added Sample images and thumbnails

kcoley:


bghgary:
Looks good.

Issues with images:
* Compatibility with extension required should not have an image since the loader should have failed with an error message.

Other notes:
* Diffuse factor in spec-gloss table has no effect because of spec factor defaulting to 1, 1, 1. The sample images make this very obvious that we should do something. 👍 
* Mesh_Primitives/Texture_Sampler table's order is reversed

We need to go over all sample images and make sure each is correct.

stevk:
Approved once the conflict is resolved. I'll remove the blank sample images along with the PR to just remove that column in compat.

Issue #371, opened 2018-03-09
Mesh_PrimitiveMode is missing some tests along Points and Lines


stevk:
- [ ] We need more points so that the model is visible. Outline a square with points? All have the same normal.
- [ ] Test points with normals
- [ ] test points w/o normals

Issue #369, opened 2018-03-08
Mesh_Indices should be Mesh_PrimitiveMode


kcoley:
change the name

PR #375, opened 2018-03-13
Make Mesh_Indices into Mesh_PrimitiveMode and make Points mode more visible

stevk:
- Renamed Mesh_Indices to Mesh_PrimitiveMode
- Added additional vertexes to the Points model, to add visibility

Fix for #369 and partial fix for #371

[Mesh_PrimitiveMode](https://github.com/stevk/glTF-Asset-Generator/blob/Meshindices/Output/Mesh_PrimitiveMode/README.md)


stevk:
@bghgary Updates have been made. Points is much more visible now, very close to lines.

PR #376, opened 2018-03-13
Change Mesh_Primitives to test having two primitives with different materials

stevk:
Changes this model group to only set the Base Color Factor attribute, and add in some models where the primitives are assigned different materials.

Fix for #370 

[Mesh_Primitives](https://github.com/stevk/glTF-Asset-Generator/blob/twomats/Output/Mesh_Primitives/README.md)


bghgary:
This table is very confusing to me. I think you should put the two materials on the top and the table should only have two columns that indicate which material is used.

stevk:
@bghgary The table has been simplified. Are there any models you want added?

bghgary:
Update baseColorFactor values to [0, 1, 0, 1] and [0, 0, 1, 1]. No need for alpha and no need for fractions.

stevk:
@bghgary Updates made.

PR #391, opened 2018-03-21
Moving textures out of figures directory

kcoley:
Below is the new texture structure:
- Textures
- Figures
    SampleImages
    Thumbnails


Source/ModelGroups/Material.cs
```diff
@@ -13,16 +13,16 @@ public Material(List<string> figures) : base(figures)
             noPrerequisite = false;
             Runtime.Image emissiveTexture = new Runtime.Image
             {
-                Uri = figures.Find(e => e.Contains("Emissive_Plane"))
+                Uri = figures.Find(e => e.Contains("Textures/Emissive_Plane")).Replace("Resources/", "")
```
stevk: Is the extra code needed? To my understanding, the previous version of this line should work just as well.

stevk:
No changes to` SampleImageHelper.ps1` ? I'd be surprised if it still worked after changing the folder structure like this.

Source/ModelGroups/Material.cs
kcoley: I moved the textures and figures into a subdirectory within the Visual studio solution:
- Resources
  - Textures
  - Figures

In changing this, the figures.Find line now locates the thumbnail images first instead of the actual textures.  To prevent that from happening, I pre-pended the Textures directory to the texture name

kcoley:
I haven't changed `SampleImageHelper.ps1` yet.  I used my `copyOver` script to move the files into the expected locations for now.

Source/ModelGroups/Material.cs
stevk: What about `.Replace("Resources/", "")`? That's just removing the "Resources/" part of the string, but `.Contains` won't care about that, since it is checking for a substring.

kcoley: The contains will get the url of the resource, for example `Resources\Textures\Emissive_Plane.png`
But we don't want to include the Resource directory as part of the destination image path, so that gets stripped out so it becomes `Textures\Emissive_Plane.png`.

stevk: Ah, I see. I missed that double `))` and though it was working on `Contains` not the result. Never mind then.,

stevk:
Ok. I won't block this PR based on that, but we'll need to get a working script in asap. Defiantly before v0.5

kcoley:
Just made an update for the powershell script to work.

PR #393, opened 2018-03-23
Update SampleImageHelper.ps1 to work with recent commits

stevk:
- Update to use the new folder layout

Quick change to get the script into a working state. We have plans to replace this within the assetGenerator in a future update, so I didn't spend time on making this script easier to maintain.

Note that the screenshots of Material_Mixed_01 and 02 were both updated. We believe this is due to a lighting change in the screenshotGenerator.


stevk:
Looks like there was an even newer version of the screenshot generator that doesn't need this fix.

Issue #394, opened 2018-03-26
Refactor generator logic


kcoley:
I made a prototype for refactoring the generator logic in json (https://github.com/kcoley/glTF-Asset-Generator/blob/structurePrototype/Source/Resources/ModelGroups/MaterialModelGroup.json)

As discussed offline, we will do the refactoring as a class that has the essential behavior of the json.

stevk:
Fixed in #402

PR #400, opened 2018-04-03
Use Directory functions instead of Reflection to generate lists of images

stevk:
Progress made on #394

This check-in stops using reflection to find images that have been included as imbedded resources, and instead queries the directory directly for the list of images.

While making this change I cleaned up a lot of the code that handled copying the images, removing the need for regex string replacements and other complicated functions. I also removed the override sample images, as we currently aren't using them


Source/FileHelper.cs
```diff
@@ -37,133 +37,80 @@ public static void ClearOldFiles(string executingAssemblyFolder, string assetFol
...
 
             // Copy those thumbnails to the destination directory
-            CopyImageFiles(executingAssembly, outputFolder, usedThumbnailImages, destinationPath);
+            CopyImageFiles(executingAssembly, outputFolder, usedThumbnailImages);
+        }
+
+        /// <summary>
+        /// Replaces '\\' with '/', for use in building a URI for an image
+        /// </summary>
+        static string FormatForUIR(string path)
```
bghgary: spelling.. Preferably using camel casing as is standard in C#: FormatForUri

Source/FileHelper.cs
```diff
@@ -37,133 +37,80 @@ public static void ClearOldFiles(string executingAssemblyFolder, string assetFol
...
+        {
+            return path.Replace(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
```
bghgary: This code is technically wrong. On platforms that use `/` by default, this will switch the `/` to `\` which is not what you want for a Uri. I would recommend using the literal characters here. Same goes for `FormatForFilesystem`. You should probably just use the `Uri` class to do this perhaps. https://stackoverflow.com/questions/1546419/convert-file-path-to-a-file-uri

Source/FileHelper.cs
```diff
@@ -37,133 +37,80 @@ public static void ClearOldFiles(string executingAssemblyFolder, string assetFol
...
+        /// </summary>
+        static void CopyThumbnailImageFiles(Assembly executingAssembly, string outputFolder, List<Runtime.Image> usedImages)
         {
             // Use the list of images to infer the list of thumbnails
-            List<Runtime.Image> usedThumbnailImages = new List<Runtime.Image>();
-            Regex changePath = new Regex(@".*\/(.*)");
-            Regex changeDestination = new Regex(@"(.+)(?=\\)");
-
-            usedThumbnailImages = DeepCopy.CloneObject(usedImages);
+            List<Runtime.Image> usedThumbnailImages = DeepCopy.CloneObject(usedImages);
```
bghgary: Why use `DeepCopy.CloneObject`?

Source/FileHelper.cs
```diff
@@ -37,133 +37,80 @@ public static void ClearOldFiles(string executingAssemblyFolder, string assetFol
...
+        /// </summary>
+        static string FormatForUIR(string path)
+        {
+            return path.Replace(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+        }
+
+        /// <summary>
+        /// Replaces '/' with '\\', for use in converting a UR back into a useable local path for an image
+        /// </summary>
+        static string FormatForFilesystem(string path)
```
bghgary: nit: Filesystem -> FileSystem

Source/FileHelper.cs
```diff
@@ -37,133 +37,80 @@ public static void ClearOldFiles(string executingAssemblyFolder, string assetFol
...
-                        if (stream == null)
-                        {
-                            throw new ArgumentException("No such image", image.Uri);
-                        }
-                        using (Stream output = File.OpenWrite(imageDestinationPath))
-                        {
-                            stream.CopyTo(output);
-                        }
-                    }
+                    var source = Path.Combine(Directory.GetCurrentDirectory(), "Resources", name);
```
bghgary: > `Directory.GetCurrentDirectory()`

Should not use this. Instead pass in the directory to use and get it from the executing assembly. No need to pass in the executing assembly anymore.

Source/FileHelper.cs
stevk: Not really needed, saved me maybe a line or two of code since I need to change every uri regardless. The code makes the list while it edits the uri values now..

Source/FileHelper.cs
stevk: I've switched from 'AltDirectorySeperator' to just the literal character. Looking into the 'Uri' class, it doesn't do what we need, since we need the literal string for 'Runtime.Uri' and the 'Uri' class handles paths in a windows specific way.

https://coding.abel.nu/2014/10/beware-of-uri-tostring/

Source/FileHelper.cs
stevk: Done

Source/FileHelper.cs
stevk: Done

Source/FileHelper.cs
stevk: Done.

Source/FileHelper.cs
```diff
@@ -37,133 +37,84 @@ public static void ClearOldFiles(string executingAssemblyFolder, string assetFol
...
+                    });
             }
 
             // Copy those thumbnails to the destination directory
-            CopyImageFiles(executingAssembly, outputFolder, usedThumbnailImages, destinationPath);
+            CopyImageFiles(executingAssemblyFolder, outputFolder, usedThumbnailImages);
+        }
+
+        /// <summary>
+        /// Replaces '\\' with '/', for use in building a URI for an image
```
bghgary: Comment is not correct anymore

Source/FileHelper.cs
```diff
@@ -37,133 +37,84 @@ public static void ClearOldFiles(string executingAssemblyFolder, string assetFol
...
+        /// </summary>
+        static string FormatForUri(string path)
+        {
+            return path.Replace(Path.DirectorySeparatorChar, '/');
+        }
+
+        /// <summary>
+        /// Replaces '/' with '\\', for use in converting a UR back into a useable local path for an image
```
bghgary: Ditto... Also UR should URI

PR #401, opened 2018-04-09
Remove automatic combos from the Material modelgroup

stevk:
This PR is more to show what I have in mind. Progress on issue #394 

There is no sorting being done or automatic combos on the Material modelgroup with this change. Once this has been implemented to all of the model groups I'll be removing the ComboHelper class entirely.

### DIFF
--- a/Output/Material/Material_06.gltf
+++ b/Output/Material/Material_06.gltf
@@ -36,7 +36,7 @@
     "generator": "glTF Asset Generator",
     "version": "2.0",
     "extras": {
-      "Attributes": "EmissiveFactor - EmissiveTexture"
+      "Attributes": "EmissiveTexture - EmissiveFactor"
     }
   },
   "buffers": [
--- a/Source/ModelGroups/Material.cs
+++ b/Source/ModelGroups/Material.cs
@@ -11,21 +11,23 @@ namespace AssetGenerator.ModelGroups
             modelGroupName = ModelGroupName.Material;
             onlyBinaryProperties = false;
             noPrerequisite = false;
+
             Runtime.Image emissiveTexture = new Runtime.Image
             {
-                Uri = imageList.Find(e => e.Contains("Emissive_Plane"))
+                Uri = imageList.Find(e => e.Contains("Textures/Emissive_Plane")).Replace("Resources/", "")
             };
             Runtime.Image normalTexture = new Runtime.Image
             {
-                Uri = imageList.Find(e => e.Contains("Normal_Plane"))
-    };
+                Uri = imageList.Find(e => e.Contains("Textures/Normal_Plane")).Replace("Resources/", "")
+            };
             Runtime.Image occlusionTexture = new Runtime.Image
             {
-                Uri = imageList.Find(e => e.Contains("Occlusion_Plane"))
-};
+                Uri = imageList.Find(e => e.Contains("Textures/Occlusion_Plane")).Replace("Resources/", "")
+            };
             usedTextures.Add(emissiveTexture);
             usedTextures.Add(normalTexture);
             usedTextures.Add(occlusionTexture);
+
             List<Vector3> planeNormals = new List<Vector3>()
             {
                 new Vector3( 0.0f, 0.0f,1.0f),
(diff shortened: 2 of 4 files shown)


Output/Material/Material_06.gltf
```diff
@@ -36,7 +36,7 @@
     "generator": "glTF Asset Generator",
     "version": "2.0",
     "extras": {
-      "Attributes": "EmissiveFactor - EmissiveTexture"
+      "Attributes": "EmissiveTexture - EmissiveFactor"
```
bghgary: Is this useful? I have never had any use for this.

Source/ModelGroups/Material.cs
```diff
@@ -11,89 +11,101 @@ public Material(List<string> imageList) : base(imageList)
             modelGroupName = ModelGroupName.Material;
             onlyBinaryProperties = false;
             noPrerequisite = false;
+
             Runtime.Image emissiveTexture = new Runtime.Image
             {
-                Uri = imageList.Find(e => e.Contains("Emissive_Plane"))
+                Uri = imageList.Find(e => e.Contains("Textures/Emissive_Plane")).Replace("Resources/", "")
```
bghgary: Should the replace be happening when populating the image list?

Source/ModelGroups/Material.cs
```diff
@@ -11,89 +11,101 @@ public Material(List<string> imageList) : base(imageList)
...
+            Property scale = new Property(Propertyname.Scale, 10.0f, Propertyname.NormalTexture);
+            Property occlusionTextureProperty = new Property(Propertyname.OcclusionTexture, occlusionTexture);
+            Property strength = new Property(Propertyname.Strength, 0.5f, Propertyname.OcclusionTexture);
+            Property emissiveTextureProperty = new Property(Propertyname.EmissiveTexture, emissiveTexture);
+            Property emissiveFactor = new Property(Propertyname.EmissiveFactor, new Vector3(1.0f, 1.0f, 1.0f));
+
+            // Declares the list of properties. The order here determins the column order.
+            properties = new List<Property>
+            {
+                normalTextureProperty,
```
bghgary: naming consistency, would recommend not putting property:

normalTexture,
normalTextureScale,
occlusionTexture,
etc.

Source/ModelGroups/Material.cs
```diff
@@ -11,89 +11,101 @@ public Material(List<string> imageList) : base(imageList)
...
+                scale,
+                occlusionTextureProperty,
+                strength,
+                emissiveTextureProperty,
+                emissiveFactor
+            };
+
+            // Declares the combos. The order here determines the number for the model (assending order) and the order that the properties are applied to a model.
```
bghgary: This is much easier to understand and read. I like it.

Source/ModelGroups/Material.cs
```diff
@@ -11,89 +11,101 @@ public Material(List<string> imageList) : base(imageList)
...
+            combos.Add(new List<Property>()
```
bghgary: In terms for formatting, I would recommend not initializing the combos member and using the collection initializer for creating lists:
```
combos = new List<List<Property>>
{
    new List<Property> { normalTexture },
    new List<Property> { occlusionTexture },
}
```

Output/Material/Material_06.gltf
stevk: It was useful when I was still working out the issues with the readme. I haven't needed it in a while. It could still be useful for someone to at-a-glance see what is being tested on a model, but I'm ok with removing this.

Source/ModelGroups/Material.cs
stevk: Oops, this change was the result of a bad rebase. ` Uri = imageList.Find(e => e.Contains("Emissive_Plane")) ` is correct

Source/ModelGroups/Material.cs
stevk: The `image` for the textures are already using that explicit name, so I added "property" to those. Suggestions on what to name the textures to differentiate them from the property?

Source/ModelGroups/Material.cs
stevk: Seems reasonable to me.

PR #402, opened 2018-04-11
ModelGroup Refactor

stevk:
Model's are now tracked as their own objects, and are created (along with their gltf data) when the modelgroup object is created. Streamlined a lot of the code related to the models and modelgroups.


Source/ModelGroup.cs
```diff
@@ -1,37 +1,146 @@
...
+            specialProperties = initialValues.specialProperties;
+            combos = initialValues.combos;
+            usedTextures = initialValues.usedTextures;
+            usedFigures = initialValues.usedFigures;
+            id = initialValues.id;
+            noSampleImages = initialValues.noSampleImages;
         }
 
-        public virtual List<List<Property>> ApplySpecialProperties(ModelGroup modelGroup, List<List<Property>> combos)
+        public Runtime.GLTF SetModelAttributes(Runtime.GLTF wrapper, Runtime.Material material, List<Property> combo, ref glTFLoader.Schema.Gltf gltf)
```
kcoley: Do you even need `ref glTFLoader.Schema.GLtf gltf` here?

Source/ModelGroup.cs
```diff
@@ -1,37 +1,146 @@
...
+            {
+                property.value(wrapper);
+            }
+
+            return wrapper;
+        }
+
+        public void PostRuntimeChanges(List<Property> combo, ref glTFLoader.Schema.Gltf gltf)
+        {
+
```
kcoley: Is this placeholder code?

Source/ModelGroup.cs
```diff
@@ -1,37 +1,146 @@
...
         }
 
-        public virtual void PostRuntimeChanges(List<Property> combo, ref glTFLoader.Schema.Gltf gltf)
+        internal Runtime.GLTF ApplyModelProperty(Runtime.GLTF wrapper, List<Property> combo)
+        {
+
```
kcoley: Is this placeholder code?

Source/ModelGroup.cs
```diff
@@ -1,37 +1,146 @@
...
+        public List<Property> requiredProperty;
+        public List<Property> specialProperties;
+        public List<List<Property>> combos;
+        public List<Runtime.Image> usedTextures;
+        public List<Runtime.Image> usedFigures;
+        public int id;
+        public bool noSampleImages;
 
-        public ModelGroup(List<string> figures)
+        public ModelGroup(InitialiseModelGroup initialValues)
```
kcoley: Can `InitialiseModelGroup` be renamed to `InitializeModelGroup` ?

Source/ModelGroups/InitialiseModelGroup.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AssetGenerator
+{
+    internal class InitialiseModelGroup
```
kcoley: Rename to `InitializeModelGroup`

Source/Program.cs
```diff
@@ -21,31 +21,41 @@ private static void Main(string[] args)
...
-                {
-                    ConstructorInfo ctor = type.GetConstructor(new Type[] { typeof(List<string>) });
-                    dynamic modelGroup = ctor.Invoke(new dynamic[] { imageList });
-                    allModelGroups.Add(modelGroup);
-                }
-            }
+                new ModelGroup(new Material(imageList)),
+            };
+
+            // Uses Reflection to create a list containing one instance of each group of models
```
kcoley: Can this be deleted?

Source/Program.cs
```diff
@@ -21,31 +21,41 @@ private static void Main(string[] args)
...
+            //        dynamic modelGroup = ctor.Invoke(new dynamic[] { imageList });
+            //        allModelGroups.Add(modelGroup);
+            //    }
+            //}
 
             var modelGroupIndex = 0;
             foreach (var modelGroup in allModelGroups)
             {
-                List<List<Property>> combos = ComboHelper.AttributeCombos(modelGroup);
+                // Creates the combos if the model group is still using automatic combos
```
kcoley: Same here, can the commented out code be deleted?

Source/Program.cs
```diff
@@ -54,20 +64,20 @@ private static void Main(string[] args)
...
                     var asset = new Runtime.Asset
                     {
                         Generator = "glTF Asset Generator",
                         Version = "2.0",
-                        Extras = new Runtime.Extras
-                        {
-                            // Inserts a string into the .gltf containing the properties that are set for a given model, for debug.
-                            Attributes = String.Join(" - ", name)
-                        }
+                        //Extras = new Runtime.Extras
```
kcoley: Same here, delete the commented code

Source/ReadmeBuilder.cs
```diff
@@ -97,15 +97,15 @@ public void SetupHeader(ModelGroup test)
...
-                if (test.properties[i].prerequisite != Propertyname.Undefined && test.properties[i].propertyGroup == 0)
-                {
-                    attributeName = test.properties[i].prerequisite.ToString() + test.properties[i].name.ToString();
-                }
-                else
-                {
-                    attributeName = test.properties[i].name.ToString();
-                }
-                attributeName = ReadmeStringHelper.GenerateNameWithSpaces(attributeName);
+                //if (test.properties[i].prerequisite != Propertyname.Undefined && test.properties[i].propertyGroup == 0)
```
kcoley: Remove the commented out code

Source/ReadmeStringHelper.cs
```diff
@@ -42,46 +42,46 @@ public static string ConvertTestValueToString(Property param)
...
                     // Generates a name for nonBinary attributes
-                    if (param.propertyGroup > 0)
-                    {
-                        output = GenerateNonbinaryName(param.name.ToString());
-                    }
-                    else
-                    {
-                        output = ":white_check_mark:";
-                    }
+                    //if (param.propertyGroup > 0)
```
kcoley: Remove commented out code

Source/ModelGroup.cs
stevk: Not for the Material model group. I thought it was required for one of the later model groups, but I'm not seeing which one at first glance. I'll remove it for now and we can re-add it if/when the need comes up.

Source/ModelGroup.cs
stevk: Placeholder. I'll likely need to do something with this to make the Compatibility model group work.

Source/ModelGroup.cs
stevk: I believe this is a dead-end that I haven't gotten around to removing yet.

Source/ModelGroup.cs
stevk: done

Source/ModelGroups/InitialiseModelGroup.cs
stevk: done

Source/Program.cs
stevk: `new ModelGroup(new Material(imageList)),` ? No, this is required. What is wrong with it?

Source/Program.cs
stevk: Like I said in the description, I've not done this yet. I want to make sure the structure is what we want before I put a lot of effort into cleaning everything up.

Source/Program.cs
stevk: see previous

Source/ReadmeBuilder.cs
stevk: see previous

Source/ReadmeStringHelper.cs
stevk: see previous

Output/Manifest.json
```diff
@@ -1,129 +1,4 @@
 [
-  {
```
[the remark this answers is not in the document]
stevk: This was a mistake. I'll make sure it gets back in when I do a pass on the readme class.

stevk:
I've made an update. I haven't updated the readme or manifest code yet, so ignore those files. 
I've gotten as close to @bghgary's code sample as I could, but I don't think I quite captured his intent. Overall it feels more complicated, but that might just be from an initial implementation standpoint.

Output/Manifest.json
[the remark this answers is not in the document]
stevk: Ok, so I figured it out. The manifest is built from scratch each time the AssetGenerator is run. Since only the Material model group is being generated, that's the only thing being added to the manifest. This issue will go away on its own as more model groups are refactored.

Source/ModelGroups/Material.cs
```diff
@@ -1,162 +1,163 @@
...
-                Uri = imageList.Find(e => e.Contains("Normal_Plane"))
-    };
-            Runtime.Image occlusionTexture = new Runtime.Image
+                var properties = new List<Property>();
+                var meshPrimitive = MeshPrimitive.CreateSinglePlane();
+                meshPrimitive.Material = new Runtime.Material();
+                meshPrimitive.Material.MetallicRoughnessMaterial = new Runtime.PbrMetallicRoughness();
+
+                // Apply the common properties to the gltf.
+                SetMetallicFactor(new List<Property>(), meshPrimitive.Material.MetallicRoughnessMaterial);
```
bghgary: * readme is out of order
* normals should be grouped with normal texture function
* separate the common properties function

Source/ModelGroups/ModelGroup.cs
```diff
@@ -0,0 +1,158 @@
...
+            SortPropertiesList(CommonProperties);
+            SortPropertiesList(Properties);
+        }
+
+        protected void SortPropertiesList(List<Property> properties)
+        {
+            // Sorts the list so every readme has the same column order, determined by enum value.
+            if (properties.Count > 0)
+            {
+                properties.Sort(delegate (Property x, Property y)
```
bghgary: can use short hand (=>) instead of delegate

Source/Property.cs
```diff
@@ -7,167 +7,25 @@ namespace AssetGenerator
 {
     public class Property
```
bghgary: all classes should be internal and the internal properties should be public.

Source/Property.cs
```diff
@@ -7,167 +7,25 @@ namespace AssetGenerator
...
     {
-        public Propertyname name { get; }
-        public dynamic value; // Could be a float, array of floats, string, or enum
-        public Propertyname prerequisite = Propertyname.Undefined;
-        public int propertyGroup;
+        internal PropertyName name;
+        internal string readmeColumnName;
```
bghgary: consider deriving the column name from name or use a dictionary to map the names

Source/ModelGroups/ModelGroup.cs
```diff
@@ -0,0 +1,158 @@
...
+
+    internal struct Model
+    {
+        public List<Property> Properties { get; set; }
+        public Runtime.GLTF GLTF;
+    }
+
+    public enum ModelGroupName
+    {
+        Undefined,
```
bghgary: is this necessary?

Source/ModelGroups/ModelGroup.cs
```diff
@@ -0,0 +1,158 @@
...
+        Mesh_PrimitivesUV,
+        Node_NegativeScale,
+        Node_Attribute,
+        Texture_Sampler,
+        Primitive_VertexColor,
+    }
+
+    public enum PropertyName
+    {
+        Undefined,
```
bghgary: ditto

Source/ModelGroups/ModelGroup.cs
```diff
@@ -0,0 +1,158 @@
...
+        Mesh_PrimitiveMode,
+        Mesh_Primitives,
+        Mesh_PrimitivesUV,
+        Node_NegativeScale,
+        Node_Attribute,
+        Texture_Sampler,
+        Primitive_VertexColor,
+    }
+
+    public enum PropertyName
```
bghgary: move to Property class?

Source/Property.cs
```diff
@@ -7,167 +7,25 @@ namespace AssetGenerator
 {
     public class Property
     {
-        public Propertyname name { get; }
-        public dynamic value; // Could be a float, array of floats, string, or enum
-        public Propertyname prerequisite = Propertyname.Undefined;
-        public int propertyGroup;
+        internal PropertyName name;
```
bghgary: Should be upper camel case

Source/ModelGroups/ModelGroup.cs
stevk: Removed

Source/ModelGroups/ModelGroup.cs
stevk: Removed

Source/ModelGroups/ModelGroup.cs
stevk: Correction to that, we are using `ModelGroupName.Undefined` as a null value when assigning camera parameters in the manifest.

Source/ModelGroups/ModelGroup.cs
stevk: Fixed. Also switched to using `CompareTo`

Source/Property.cs
stevk: Fixed

Source/ModelGroups/ModelGroup.cs
stevk: Fixed

Source/Property.cs
stevk: I think I fixed all of these. I likely missed a few, and there are probably things that could be set as `protected` that haven't been yet.

Source/Property.cs
stevk: Fixed here and elsewhere.

Source/AssetGenerator.csproj
```diff
@@ -13,9 +13,9 @@
     <None Remove="Images\**" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Remove="Attributes.cs" />
```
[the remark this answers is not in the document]
stevk: It is not in source. Verified in both Visual Studio and File Explorer.

Source/Manifest.cs
```diff
@@ -3,81 +3,81 @@
...
+                cameratranslation.CopyTo(Translation);
             }
         }
 
         // Used to track camera properties for model groups that need a custom camera
-        internal static class CustomCameraList
+        protected static class CustomCameraList
         {
-            internal static List<ModelCameraPairing> customCameraList;
+            static List<ModelCameraPairing> CustomCameras;
```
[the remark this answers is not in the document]
stevk: Fixed

Source/Manifest.cs
```diff
@@ -3,81 +3,81 @@
...
             public Camera(Vector3 cameratranslation)
             {
-                cameratranslation.CopyTo(translation);
+                cameratranslation.CopyTo(Translation);
             }
         }
 
         // Used to track camera properties for model groups that need a custom camera
-        internal static class CustomCameraList
+        protected static class CustomCameraList
```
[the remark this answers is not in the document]
stevk: Fixed

Output/Material/Manifest.json
```diff
@@ -1,88 +1,88 @@
 {
-  "folder": "Material",
-  "models": [
+  "Folder": "Material",
```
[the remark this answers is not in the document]
stevk: Fixed

Source/Manifest.cs
```diff
@@ -46,9 +45,9 @@ public Camera(Vector3 cameratranslation)
         }
 
         // Used to track camera properties for model groups that need a custom camera
-        protected static class CustomCameraList
+        private static class CustomCameraList
         {
-            static List<ModelCameraPairing> CustomCameras;
+            private static List<ModelCameraPairing> CustomCameras;
```
bghgary: private variables should be lower camel case

Source/Program.cs
```diff
@@ -17,6 +17,14 @@ private static void Main(string[] args)
             var pathSeparator = Path.DirectorySeparatorChar;
             string outputFolder = Path.GetFullPath(Path.Combine(executingAssemblyFolder, String.Format(@"..{0}..{0}..{0}..{0}Output", pathSeparator)));
             List<Manifest> manifestMaster = new List<Manifest>();
+            var jsonSerializer = new Newtonsoft.Json.JsonSerializer
+            {
+                Formatting = Newtonsoft.Json.Formatting.Indented,
+                ContractResolver = new Newtonsoft.Json.Serialization.CamelCasePropertyNamesContractResolver
+                {
```
bghgary: replace {} with () on a single line since there are no properties to set

Source/Property.cs
```diff
@@ -27,6 +27,26 @@ public Property(PropertyName enumName, object displayValue, int group = 0)
         }
     }
 
+    internal class PropertyComparer : IEqualityComparer<Property>
```
bghgary: Where is this used?

bghgary: You should probably just override the methods on the Property class itself instead of doing this.

Source/Program.cs
stevk: Fixed. Good catch

Source/Property.cs
stevk: Used in ModelGroup.cs, line 83. Good point, I've changed this to be an override.

stevk:
Looks like the order between a couple of the model groups changed. I think this is a difference in how Visual Studio orders alphabetically vs GitHub. I think we should leave this change in place, since the order will now better match the order of the folders as displayed on GitHub.

Mesh_PrimitiveVertexColor 13 -> 11
Mesh_Primitives 11 -> 12
Mesh_PrimitivesUV 12 -> 13

bghgary:
That's fine. Looks like GitHub sorts case sensitive.

PR #415, opened 2018-05-21
Remove the Id property from ModelGroup

### LINKED ISSUE #413: Id property in ModelGroup is not intuitive and likely not needed
We could remove `Id` completely, and instead use the enum value of `ModelGroupName`.

Otherwise `Id` should be renamed to something more descriptive, like `ReadmeIndex`.

stevk:
Changes made to the manifest code and ModelGroupName enum were done to keep the existing order intact. Also removed a String.Format that hadn't been updated yet.

Fixes #413

### DIFF
--- a/Source/Manifest.cs
+++ b/Source/Manifest.cs
@@ -9,7 +9,7 @@ namespace AssetGenerator
         public List<Model> Models = new List<Model>();
 
         // Model group, to be listed in the manifest as the folder name
-        public Manifest(ModelGroupName name)
+        public Manifest(ModelGroupId name)
         {
             Folder = name.ToString();
         }
@@ -22,7 +22,7 @@ namespace AssetGenerator
             public string SampleImageName;
             public Camera Camera;
 
-            public Model(string name, ModelGroupName modelGroupName, bool noSampleImages)
+            public Model(string name, ModelGroupId modelGroupName, bool noSampleImages)
             {
                 FileName = name;
                 if (noSampleImages == false)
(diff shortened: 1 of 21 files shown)


Source/ModelGroup.cs
```diff
@@ -13,7 +13,6 @@ internal abstract partial class ModelGroup
         public List<Property> Properties { get; private set; }
         public List<Runtime.Image> UsedTextures { get; private set; }
         public List<Runtime.Image> UsedFigures { get; private set; }
-        public int Id { get; set; }
```
bghgary: I think we should consider changing `ModelGroupName` to `ModelGroupId` and change the property `Name` to `Id`

Source/ReadmeBuilder.cs
```diff
@@ -121,7 +121,7 @@ public void SetupTable(ModelGroup test, int modelIndex, List<Property> model)
         {
             string modelGroupName = test.Name.ToString();
             string modelNumber = modelIndex.ToString("D2");
-            string liveURL = string.Format("https://bghgary.github.io/glTF-Assets-Viewer/?folder={0}&model={1}", test.Id, modelIndex);
+            string liveURL = $"https://bghgary.github.io/glTF-Assets-Viewer/?folder={Convert.ToInt32(test.Name)}&model={modelIndex}";
```
bghgary: `{Convert.ToInt32(test.Name)}` can be `{test.Name:D}`

Source/ModelGroup.cs
stevk: I thought the term `Id` was somewhat ambiguous in this context, and the enum is used as a string in quite a few more places than it is used as an int. Is there a strong argument for `Id`?

bghgary: The word `name` is confusing to me as an enum. `name` is typically a string.

Source/Manifest.cs
```diff
@@ -52,16 +52,16 @@ private static class CustomCameraList
             public class ModelCameraPairing
             {
                 public Camera camera;
-                public ModelGroupName modelGroup;
+                public ModelGroupId? modelGroup;
 
-                public ModelCameraPairing(Camera cameraSettings, ModelGroupName name)
+                public ModelCameraPairing(Camera cameraSettings, ModelGroupId? name)
```
bghgary: nit: name -> modelGroup

Source/ReadmeBuilder.cs
```diff
@@ -161,7 +161,7 @@ public void SetupTable(ModelGroup test, int modelIndex, List<Property> model)
         public void WriteOut(Assembly executingAssembly, ModelGroup test, string assetFolder)
         {
             string template;
-            string templatePath = $"AssetGenerator.ReadmeTemplates.{test.Name.ToString()}.md";
+            string templatePath = $"AssetGenerator.ReadmeTemplates.{test.Id.ToString()}.md";
```
bghgary: `.ToString()` is extraneous

stevk: Fixed

PR #417, opened 2018-05-21
Rename VertexUV0 to TextureCoordinateComponentType

### LINKED ISSUE #408: Mesh_PrimitiveAttribute 'Vertex UV 0' should just be 'Vertex UV'
There is only one vertex in this model group, so the '0' isn't needed. This is likely leftover from when there were multiple UVs.

stevk:
There was no need for the 0 in either of these model groups. 'TextureCoordinate' is more accurate to the documentation, in my opinion, as well.

Fixes #408

### DIFF
--- a/Output/Buffer_Interleaved/README.md
+++ b/Output/Buffer_Interleaved/README.md
@@ -11,7 +11,7 @@ The following table shows the properties that are set for every model.
 
 The following table shows the properties that are set for a given model.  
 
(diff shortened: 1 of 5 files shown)


stevk:
Would `VertexUV` be more acceptable? Otherwise I'm fine will killing this PR.

bghgary:
I'm okay with dropping the 0, but I don't think it's necessary. We didn't use texture coordinate before because it was too long.

PR #418, opened 2018-05-21
Fix issue with Mesh_Primitives where the readme didn't match the models

### LINKED ISSUE #411: Mesh_Primitives refering to materials as 1 and 0 in the readme is incorrect with the current models.
In the JSON for models 01 and 02 the primitives are always using the same material, the color values is all that is changed.

I think that ideally we should find a way to make Primitive 0 use material 1 and Primitive 1 use material 0.

stevk:
Without this change, the [readme for Mesh_Primitives](https://github.com/bghgary/glTF-Asset-Generator/blob/master/Output/Mesh_Primitives/README.md) inaccurately states that we're switching the index of the material used. With this change the models really do change as described.

Fixes #411


bghgary:
Was this broken even before the refactor?

stevk:
Yes, this was broken before the refactor.

stevk:
Updated per our discussion

PR #422, opened 2018-05-23
Make Points and lines asymmetrical in Mesh_PrimitiveMode 

### LINKED ISSUE #403: Primitive Mode should have asymmetrical lines and points
It is difficult to tell if the points and lines are reversed when loading

stevk:
Uses asymmetrical positions for points and lines, to help make it visually apparent if they are reversed.

Fixes #403


Source/ModelGroup.cs
```diff
@@ -101,6 +101,17 @@ public static List<Vector3> GetSinglePlanePositions()
                 };
             }
 
+            public static List<Vector3> GetSinglePlaneNonReversablePositions()
```
bghgary: Might consider not factoring this out as it's specifically for one model group.

Issue #419, opened 2018-05-22
What do we want to test instancing with


stevk:
Below is the list so far. 

- [x] Nodes pointing to the same mesh
- [x] Nodes pointing to the same skin
- [x] Primitives pointing to the same material
- [x] Materials pointing to the same texture

stevk:
@kcoley could you mark which areas were covered by #462

stevk:
There are models using instancing in several modelgroups, but there should be one modelgroup specifically showcasing instancing, to make these easier to find.

stevk:
Morph is NYI.

PR #424, opened 2018-05-24
Instancing Model Group

stevk:
Readme : https://github.com/stevk/glTF-Asset-Generator/blob/instancing/Output/Instancing/README.md 

In order to instance properties I had to change the runtime code to make some equivalency checks. This caused models in other model groups to use instancing where appropriate, but no visible changes with the models.

There are more models that need to be added to this model group (see #419) but they will required a different kind of equivalency check due to the properties being added to the buffer, which makes it harder to compare values after the fact. My thought is to add some kind of list(s) that will act in parallel with what is added to the buffer.


Output/Instancing/Instancing_00.gltf
```diff
@@ -0,0 +1,136 @@
...
+    {
+      "mesh": 0,
+      "translation": [
+        -0.5,
+        0.0,
+        0.0
+      ]
+    },
+    {
+      "mesh": 1,
```
kcoley: Looks like the mesh is not instanced here

stevk: Correct, this is an issue. I could make a short term fix by removing the indices from both meshes, otherwise this will require the same solution needed by other models with properties added to the buffer.

stevk:
Per our discussion, instead of checking models for equivalence we should implement a dictionary in the runtime layer.

stevk:
This has been covered by a commit by @kcoley

Issue #414, opened 2018-05-21
Implement tests for node animation


bghgary:
**Infrastructure**

Runtime
- [ ] Add runtime layer for animation

Screenshots (later):
- [ ] Generate animated gifs

Validation:
- [ ] Add data somewhere to store the animation data
- [ ] Figure out how to validate the data

**Tests**

For all tests:
- Add a point of reference cube with a texture that has labels for each side

Properties (using linear interpolation):
- [ ] Translation
- [ ] Rotation
- [ ] Scale

Interpolation (using translation and rotation specifically for cubic spline):
- [ ] Translation with STEP
- [ ] Translation with CUBICSPLINE
- [ ] Rotation with CUBICSPLINE

Misc:
- [ ] Cover #404 by creating one curve that doesn't start at zero

Multiple channels and samplers (maybe?):
- [ ] One model with translation, rotation, and scale

stevk:
@bghgary To do "One animation, two channels for two nodes" we'll need a model with multiple nodes. Do we want to create something specifically for this, or duplicate the original node? Duplicating will be easier, as it probably isn't worth the effort of making another model just for this one case.

bghgary:
Having two cubes seems fine to me.

donmccurdy:
A suggested addition: an animation with two channels, where one channel omits `node`. Per spec:

> When node isn't defined, channel should be ignored. Valid path names are "translation", "rotation", "scale", and "weights".

... an implementation should ignore one channel and be able to play the other.

/cc https://github.com/KhronosGroup/glTF/pull/1301#issuecomment-422057805

bghgary:
> A suggested addition: an animation with two channels, where one channel omits `node`

Sure. It might be better to open another issue for this as this issue has been merged except for the screenshots with animated gifs and the validation parts, which we might split into other issues as well.

donmccurdy:
Thanks, done: https://github.com/KhronosGroup/glTF-Asset-Generator/issues/453

PR #426, opened 2018-05-29
Node_Animation Model Group

### LINKED ISSUE #404: Make sure to cover this case when implementing animation tests
https://github.com/BabylonJS/Babylon.js/issues/4224

stevk:
Implements the Node_Animation model group.

Changed some of the runtime animation objects from structs to classes because the structs were causing issues with how we've been passing references.

Addresses several parts of #414 
Fixes #404, #428 

TODO:
- [x] I'm not sure I've correctly implemented a model for animating a curve that doesn't start at 0.
- [x] A proper texture still needs to be created to show which part of the model is supposed to be facing which way. 
- [ ] The validator shows an error in the bufferView for "Animation Sampler Output" for all of the models that are doing rotations. I'm not sure of the exact cause. "Accessor element at index # is not of unit length: 2."
- [ ] I'm sure we can pick better values for the animations that what is currently plugged in. 
- [ ] Maybe we should add a notes column to describe the expected animation for each model?


Output/Animation_Node/README.md
```diff
@@ -0,0 +1,17 @@
+These models are intended to test the basic attributes of animation on a node.  
+
+The last model on this list creates a curve which doesn't start at Zero.  
+
+The following table shows the properties that are set for a given model.  
+
+|   | Sample Image | Translation | Rotation | Scale | Interpolation |
```
bghgary: Let's try a single column for `Target(s)` and list the targets.

Source/Runtime/Animation.cs
```diff
@@ -4,9 +4,9 @@
 
 namespace AssetGenerator.Runtime
 {
-    internal struct Animation
+    internal class Animation
     {
         public String Name { get; set; }
-        public List<AnimationChannel> AnimationChannels { get; set; }
```
bghgary: Restore the `Animation` prefix for types.

Source/ModelGroups/Animation_Node.cs
```diff
@@ -0,0 +1,296 @@
...
+                properties.Add(new Property(PropertyName.Interpolation, "Cubic Spline"));
+            }
+
+            void CreateMultipleChannelsWithUniqueTargets(List<Property> properties, List<Runtime.Channel> channels, Runtime.Node node)
+            {
+                // The first channel is already added as a common property.
+                channels.Add(new Runtime.Channel());
+                channels.Add(new Runtime.Channel());
+
+                SetTranslationChannelTarget(properties, channels[0], node);
```
bghgary: create a new properties list and then combine the strings afterwards into the real properties list

Source/Runtime/Extensions/KHR_texture_transform.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Numerics;
```
bghgary: oops, i accidentally put this in. please remove this.

Source/ModelGroup_Cube.cs
```diff
@@ -0,0 +1,129 @@
...
+        protected static partial class MeshPrimitive
+        {
+            public static Runtime.MeshPrimitive CreateCube()
+            {
+                return new Runtime.MeshPrimitive
+                {
+                    Positions = new List<Vector3>()
+                    {
+                        // Left
+                        new Vector3( -0.5f, -0.5f, -0.5f),
```
bghgary: Scale to 0.1

Source/ModelGroups/Animation_Node.cs
```diff
@@ -0,0 +1,296 @@
...
+                };
+                properties.Add(new Property(PropertyName.Scale, ":white_check_mark:"));
+            }
+
+            void SetLinearSampler(List<Property> properties, Runtime.Channel channel)
+            {
+                channel.Sampler = new Runtime.LinearAnimationSampler<Vector3>(
+                    new List<float>
+                    {
+                        0.0f,
```
bghgary: 0.0, 1.0, 2.0 to all

Source/ModelGroups/Animation_Node.cs
```diff
@@ -0,0 +1,296 @@
...
+                        0.5f,
+                        1.0f,
+                    },
+                    new List<Vector3>
+                    {
+                        new Vector3(0, 0, 0),
```
bghgary: (-0.5, 0.0, 0.0), (0.0, 0.0, 0.0), (0.5, 0.0, 0.0)

Source/ModelGroup_Cube.cs
```diff
@@ -0,0 +1,129 @@
...
+                            new Vector2(0.500f, 0.000f),
+                            new Vector2(0.000f, 0.000f),
+
+                            // Front
+                            new Vector2(0.500f, 0.333f),
+                            new Vector2(1.000f, 0.333f),
+                            new Vector2(1.000f, 0.000f),
+                            new Vector2(0.500f, 0.000f),
+
+                            // Right
```
bghgary: Left <-> Right

Source/ModelGroups/Animation_Node.cs
```diff
@@ -0,0 +1,296 @@
...
+                channel.Sampler = new Runtime.LinearAnimationSampler<Quaternion>(
+                    new List<float>
+                    {
+                        0.0f,
+                        0.5f,
+                        1.0f,
+                    },
+                    new List<Quaternion>
+                    {
+                        new Quaternion(0, 0, 0, 0),
```
bghgary: Show Right, Front, then Left

Source/ModelGroups/Animation_Node.cs
```diff
@@ -0,0 +1,296 @@
...
+            {
+                channel.Target = new Runtime.ChannelTarget
+                {
+                    Node = node,
+                    Path = Runtime.ChannelTarget.PathEnum.SCALE,
+                };
+                properties.Add(new Property(PropertyName.Scale, ":white_check_mark:"));
+            }
+
+            void SetLinearSampler(List<Property> properties, Runtime.Channel channel)
```
bghgary: Create a new function for Scale and rename this to be specific for Translation. Make scale half, normal, double

Source/ModelGroups/Animation_Node.cs
```diff
@@ -0,0 +1,296 @@
...
+                        {
+                            InTangent = new Vector3(1, 1, 1),
+                            Value = new Vector3(0.1f, 0.1f, 0.1f),
+                            OutTangent = new Vector3(1,1, 1)
+                        },
+                        new Runtime.CubicSplineSampler<Vector3>.Key
+                        {
+                            InTangent = new Vector3(1, 1, 1),
+                            Value = new Vector3(0.1f, 0.1f, 0.1f),
+                            OutTangent = new Vector3(1, 1, 1)
```
bghgary: Change to zero.

Source/ModelGroups/Animation_Node.cs
```diff
@@ -0,0 +1,296 @@
...
+                        0.0f,
+                        0.5f,
+                        1.0f,
+                    },
+                    new List<Runtime.CubicSplineSampler<Vector3>.Key>
+                    {
+                        new Runtime.CubicSplineSampler<Vector3>.Key
+                        {
+                            InTangent = new Vector3(0, 0, 0),
+                            Value = new Vector3(1, 1, 1),
```
bghgary: -0.5, 0.0, 0.0

Output/Animation_Node/README.md
stevk: The `(s)` causes an error when used as a enum, so I went with `Targets` instead.

stevk: Fixed

Source/ModelGroups/Animation_Node.cs
stevk: Fixed

Source/Runtime/Animation.cs
stevk: Fixed

Source/Runtime/Extensions/KHR_texture_transform.cs
stevk: Fixed

Source/ModelGroups/Animation_Node.cs
stevk: Fixed

Source/ModelGroups/Animation_Node.cs
stevk: Fixed

Source/ModelGroups/Animation_Node.cs
stevk: Fixed

Source/ModelGroup_Cube.cs
stevk: Fixed

Source/ModelGroups/Animation_Node.cs
stevk: Fixed

Source/ModelGroup_Cube.cs
stevk: Fixed

Source/ModelGroups/Animation_Node.cs
stevk: This left a jump at the end of the animation. I believe it looks better with a full loop, so I opted for the following for a repeating Left > Right > Left motion:

```
channel.Sampler = new Runtime.LinearAnimationSampler<Vector3>(
                    new List<float>
                    {
                        0.0f,
                        2.0f,
                        4.0f,
                    },
                    new List<Vector3>
                    {
                        new Vector3(-0.5f, 0.0f, 0.0f),
                        new Vector3(0.5f, 0.0f, 0.0f),
                        new Vector3(-0.5f, 0.0f, 0.0f),
                    });
```

Source/ModelGroups/Animation_Node.cs
stevk: I made scale loop as well. Is it really nessessary to specify the 0,0,0 value? As far as I can tell, setting that middle value doesn't make a difference due to being exactly between the min/max values with equal time spacing between.

stevk:
Looking into #428 tomorrow. The model also appears to change scale when rotating, but I'm not sure if that is an issue or not.

Source/ModelGroups/Animation_Node.cs
```diff
@@ -0,0 +1,343 @@
...
+                            InTangent = new Vector3(1, 1, 1),
+                            Value = new Vector3(0.1f, 0.1f, 0.1f),
+                            OutTangent = new Vector3(1, 1, 1)
+                        }
+                    });
+
+                properties.Add(new Property(PropertyName.Interpolation, "Cubic Spline"));
+            }
+
+            void SetLinearSamplerForRotation(List<Property> properties, Runtime.AnimationChannel channel)
```
bghgary: Move this with the linear functions

Output/Animation_Node/README.md
```diff
@@ -0,0 +1,15 @@
+These models are intended to test the basic attributes of animation on a node.  
+
+The last model on this list creates a curve which doesn't start at Zero.
```
bghgary: Remove this

Source/Runtime/AnimationChannelTarget.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace AssetGenerator.Runtime
 {
-    internal struct AnimationChannelTarget
+    internal class ChannelTarget
```
bghgary: Revert this

Source/Property.cs
```diff
@@ -142,6 +142,8 @@ internal enum PropertyName
         Translation,
         Rotation,
         Scale,
+        Targets,
```
bghgary: Change to singular

stevk: Fixed

Source/Runtime/AnimationChannelTarget.cs
stevk: Fixed

Output/Animation_Node/README.md
stevk: Fixed

Source/Runtime/AnimationChannelTarget.cs
bghgary: I meant the name and not the class to struct change.

stevk: Oops, switched that back around.

PR #435, opened 2018-06-27
Refactor vertices to MeshPrimitiveVertex + Skeletons

kcoley:


Source/ModelGroups/Material_MetallicRoughness.cs
```diff
@@ -8,6 +9,19 @@ internal class Material_MetallicRoughness : ModelGroup
...
+        //    var verticesEnumerator = vertices.GetEnumerator();
+        //    var propertiesEnumerator = properties.GetEnumerator();
+
+        //    verticesEnumerator.Reset();
+        //    propertiesEnumerator.Reset();
+        //    while (verticesEnumerator.MoveNext() && propertiesEnumerator.MoveNext())
+        //    {
+        //        action(verticesEnumerator.Current, propertiesEnumerator.Current);
+        //    }
+        //}
```
stevk: Block of commented out code here. Remove it if it isn't needed?

PR #442, opened 2018-07-31
asymmetric in and out tangents

andreasplesch:
for translation cubic spline example


andreasplesch:
Looks good. Thanks for adding these.

Issue #396, opened 2018-03-29
Sample images that we intend to override are not being overridden


stevk:
Low priority, as there currently are no models that we want to override the sample image on and we intend to replace this script with code anyway. 

Looks like this code in the powershell script is comparing an image name to a file name, which will always be false.

```
            $override = $False
            if ($existingImageList.Count -gt 0)
            {
                foreach($existingImage in $existingImageList){
                    if ($existingImage -eq $model.fileName)
                    {
                        $override = $True
                    }
                }
            }
```

Also, the list of models to override isn't fully populated. Likely an issue with the list size not being increased.

PR #465, opened 2018-10-25
Add additional models to the Animation_Skin model group

### LINKED ISSUE #459: Add more tests for skin animation
- [ ]  Joint hierarchy has no common root. (KhronosGroup/glTF#1270)
As discussed, the skeleton property should probably be removed, and can already be ignored by implementations. In addition to that, we should have an example asset for which no common root is possible (e.g. bones form two subtrees starting from the scene root).
not skeleton property is kept.
 
- [ ] Multiple meshes sharing a single skin. (KhronosGroup/glTF#1285)
 
- [ ] Many (>4) joint vertex influences. (KhronosGroup/glTF#1381)

stevk:
https://github.com/stevk/glTF-Asset-Generator/tree/skinbuilder/Output/Animation_Skin

Adds three models to Animation_Skin.

- [x] MODEL 06 - Multiple meshes sharing a single skin. (KhronosGroup/glTF#1285)

- [x] MODEL 07 - Joint hierarchy has no common root. (KhronosGroup/glTF#1270)
As discussed, the skeleton property should probably be removed, and can already be ignored by implementations. In addition to that, we should have an example asset for which no common root is possible (e.g. bones form two subtrees starting from the scene root).
not skeleton property is kept.
 
- [x] MODEL 12 - Many (>4) joint vertex influences. (KhronosGroup/glTF#1381)

As a part of this PR, this also:
- Fixes SampleImageHelper so that it correctly uses provided screenshots when we want to override the generated one
- Creates a ModelGroup_Skinbuilder class that is used to dynamically generate several of the skins and mesh's used by Animation_Skin.
- Added support for setting camera position in the manifest on a per-model basis, and updated all of the skins models to take advantage of this.

Fixes #459 and #396


Source/ModelGroup_SkinBuilder.cs
```diff
@@ -0,0 +1,216 @@
...
+                        (
+                            inverseBindMatrix: Matrix4x4.CreateTranslation(new Vector3(0.0f, 0.0f, -positions[nodeIndex * 2].Z)),
+                            node: jointHierarchyNodes[nodeIndex]
+                        ));
+                    }
+                    translationMultiplier--;
+                }
+
+                // Assign weights
+                var weightsList = new List<List<Runtime.JointWeight>>();
```
bghgary: nit: rename to weights

Source/ModelGroup_SkinBuilder.cs
```diff
@@ -0,0 +1,216 @@
...
+                        Name = name,
+                        Rotation = rotationToUse,
+                        Translation = translationVector,
+                    });
+
+                    // Add as a child of the previous node
+                    jointHierarchyNodes[nodeIndex - 1].Children = new[] { jointHierarchyNodes[nodeIndex] };
+                }
+
+                var skinJointsList = new List<Runtime.SkinJoint>();
```
bghgary: nit: skinJoints (no List)

Source/ModelGroup_SkinE.cs
```diff
@@ -85,7 +85,20 @@ public static List<Runtime.Node> CreatePlaneWithSkinE()
                     },
                 };
 
-                Matrix4x4 baseRotation = Matrix4x4.CreateFromYawPitchRoll(FloatMath.ConvertDegreesToRadians(60.0f), FloatMath.ConvertDegreesToRadians(-90.0f), 0.0f);
+                // Two different ways to setup the joints and weights, where either there is only one parent or there are two parents that have the same child.
+                if (jointsHaveCommonParent == true)
```
bghgary: remove `== true`

Source/ModelGroups/Animation_Skin.cs
```diff
@@ -16,19 +16,25 @@ public Animation_Skin(List<string> imageList)
             UseFigure(imageList, "skinC");
             UseFigure(imageList, "skinD");
             UseFigure(imageList, "skinE");
+            UseFigure(imageList, "skinF");
+            var closeCameraValue = new Vector3(0.5f, 0.0f, 0.6f);
```
bghgary: rename Value to Translation

Source/ModelGroups/Animation_Skin.cs
```diff
@@ -205,7 +212,74 @@ Runtime.Animation CreateFoldingAnimation(Runtime.Node jointRootNode, List<Runtim
 
                     properties.Add(new Property(PropertyName.Description, "`skinA` where `joint1` is animated with a rotation and `joint1` has a triangle mesh attached to it."));
                 }),
-                CreateModel((properties, animations, nodes) => {
+                CreateModel((properties, animations, nodes, camera) => {
+                    // TODO WORKING ON THIS ONE
```
bghgary: remove

Source/ModelGroups/Animation_Skin.cs
```diff
@@ -205,7 +212,74 @@ Runtime.Animation CreateFoldingAnimation(Runtime.Node jointRootNode, List<Runtim
...
+                                    }
+                                }
+                            }
+                        }
+                    });
+
+                    camera.SetTranslationWithVector3 = midCameraValue;
+                    properties.Add(new Property(PropertyName.Description, "`skinA` where there are two meshes sharing a single skin."));
+                }),
+                    CreateModel((properties, animations, nodes, camera) => {
```
bghgary: indenting

Source/ModelGroups/Animation_SkinType.cs
```diff
@@ -16,11 +16,12 @@ public Animation_SkinType(List<string> imageList)
             Model CreateModel(Action<List<Property>, Runtime.MeshPrimitive> setProperties)
             {
                 var properties = new List<Property>();
-                var nodes = Nodes.CreatePlaneWithSkinA();
+                var nodes = Nodes.CreateFoldingPlaneSkin("skinA", 2, 3);
                 var animations = new List<Runtime.Animation>();
                 var meshPrimitive = nodes[0].Mesh.MeshPrimitives.First();
                 var jointComponentType = meshPrimitive.JointComponentType;
                 var weightComponentType = meshPrimitive.WeightComponentType;
+                var closeCameraValue = new Manifest.Camera(new Vector3(0.5f, 0.0f, 0.6f));
```
bghgary: closeCamera

Source/Runtime/GLTFConverter.cs
```diff
@@ -1649,28 +1649,57 @@ private glTFLoader.Schema.MeshPrimitive ConvertMeshPrimitiveToSchema(Node runtim
...
+                        totalSets = 2;
+                    }
+                }
+
+                for (int set = 0; set < totalSets; set++)
+                {
+                    var weightAccessorName = "WEIGHTS_0";
+                    var jointIndicesAccessorName = "JOINTS_0";
+                    int jointWeightIndexStartPosition = 0;
+                    if (set == 1)
```
bghgary: use string interpolation and set * 4 for start position

Source/Manifest.cs
```diff
@@ -46,20 +47,31 @@ public Model(string name, ModelGroupId modelGroupId, bool noSampleImages, Camera
...
+            public readonly Vector3 Translation = new Vector3();
 
-            public Camera(Vector3 cameraTranslation)
+            public Camera(Vector3 cameraOffset)
             {
-                cameraTranslation.CopyTo(Translation);
+                Translation = cameraOffset;
             }
         }
+        //public class Camera
```
bghgary: remove

Source/ModelGroup_Prism.cs
```diff
@@ -7,7 +7,7 @@ internal abstract partial class ModelGroup
     {
         protected static partial class Mesh
         {
-            public static Runtime.Mesh CreatePrism(Vector4 color, Vector3? Scale = null)
+            internal static Runtime.Mesh CreatePrism(Vector4 color, Vector3? Scale = null)
```
bghgary: internal isn't necessary

Source/ModelGroup_SkinBuilder.cs
```diff
@@ -80,24 +69,22 @@ public static List<Runtime.Node> CreateFoldingPlaneSkin(string skinName, int num
                     Translation = translationVectorJoint0,
                 });
                 // Create the child nodes in the joint hierarchy
+                int jointIndex1 = 1;
```
bghgary: why `jointIndex1` instead of `jointIndex`?

bghgary: you can put jointIndex in the for loop statement if scope is the issue:
`for (int nodeIndex = 1, jointIndex = 1; ...`

Source/Runtime/GLTFConverter.cs
```diff
@@ -1649,31 +1649,32 @@ private glTFLoader.Schema.MeshPrimitive ConvertMeshPrimitiveToSchema(Node runtim
             }
             if (runtimeMeshPrimitive.VertexJointWeights != null && runtimeMeshPrimitive.VertexJointWeights.Any())
             {
-                var totalSets = 1;
+                // Find the max jointweights count
+                var counts = new List<int>();
```
bghgary: There is no reason to create a list if you are just going to get the max. Just store a single int and call Math.Max

Source/Runtime/GLTFConverter.cs
```diff
@@ -1687,10 +1688,17 @@ private glTFLoader.Schema.MeshPrimitive ConvertMeshPrimitiveToSchema(Node runtim
                         var jointWeightIndexEndPosition = vertexJointWeightsCount;
                         if (vertexJointWeightsCount > 4)
                         {
-                            vertexJointWeightsCount = 4;
+                            if (vertexJointWeightsCount - jointWeightIndexStartPosition < 4)
```
bghgary: Can you explain what you are doing here?

Source/Vec3ToFloatArrayJsonConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Numerics;
+
+namespace AssetGenerator
+{
+    class Vec3ToFloatArrayJsonConverter : JsonConverter
+    {
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            Vector3? vec3 = value as Vector3?;
```
bghgary: `Vector3?` seems odd here. Does it have to be a nullable type?

Source/Vec3ToFloatArrayJsonConverter.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Numerics;
+
+namespace AssetGenerator
+{
+    class Vec3ToFloatArrayJsonConverter : JsonConverter
```
bghgary: Recommend spelling out Vector3

Source/Vec3ToFloatArrayJsonConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            Vector3? vec3 = value as Vector3?;
+            float[] floatArray = new float[3];
+            vec3.Value.CopyTo(floatArray);
+            serializer.Serialize(writer, floatArray);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            throw new NotImplementedException();
```
bghgary: Should add 
```C#
    public override bool CanRead
    {
        get { return false; }
    }
```

Source/ModelGroup_SkinBuilder.cs
stevk: Scope was the issue. I had no idea that could be done, thanks!

Source/Runtime/GLTFConverter.cs
stevk: If there are more than 4 jointweights, Check if we're currently iterating on the last jointweights (e.g. jointweights 5-7 of 7). Otherwise iterate on the jointweights list total count.

If we are iterating on the end of the jointweights list, then iterate only the number of jointweights left (number in the range of 1 to 4).

If we're not at the end of the jointweight list (which is greater than 4 in size) then iterate on a full set of 4 jointweights.

I'll add the following comment

```
// Determines how many jointweights will be iterated on during this set.
// If four or less, just use the jointweights count.
// Otherwise use 4 if we're not within 4 of the end of the jointweight list count.
// When the value is near the end of the list, use only up to the remaining number (In the range of  1 to 4)
```

Source/ModelGroup_SkinBuilder.cs
```diff
@@ -125,8 +123,7 @@ internal static List<Runtime.Node> CreateFoldingPlaneSkin(string skinName, int n
 
                 // Assign weights
                 var weights = new List<List<Runtime.JointWeight>>();
-                int jointIndex2 = 0;
-                for (int vertexPairIndex = 0; vertexPairIndex < numberOfVertexPairs; vertexPairIndex++)
+                for (int vertexPairIndex = 0, jointIndex2 = 0; vertexPairIndex < numberOfVertexPairs; vertexPairIndex++)
```
bghgary: jointIndex2 -> jointIndex

Source/Runtime/GLTFConverter.cs
```diff
@@ -1650,13 +1650,13 @@ private glTFLoader.Schema.MeshPrimitive ConvertMeshPrimitiveToSchema(Node runtim
             if (runtimeMeshPrimitive.VertexJointWeights != null && runtimeMeshPrimitive.VertexJointWeights.Any())
             {
                 // Find the max jointweights count
-                var counts = new List<int>();
+                //var counts = new List<int>();
```
bghgary: remove

Source/Runtime/GLTFConverter.cs
```diff
@@ -1684,6 +1684,10 @@ private glTFLoader.Schema.MeshPrimitive ConvertMeshPrimitiveToSchema(Node runtim
                     var normalized = false;
                     foreach (var vertexJointWeights in runtimeMeshPrimitive.VertexJointWeights)
                     {
+                        // Determines how many jointweights will be iterated on during this set.
```
bghgary: Let's go over this in person. I still don't know what this means.

PR #469, opened 2018-11-13
Hide the model that has > 4 jointweights per vertex

stevk:
None of the clients that have setup previews work with this model. Hiding it for now, since the spec doesn't have an explicit direction on how this case should be handled.

- [x] Comment out model code
- [x] Comment out figure for this skin in the readme
- [x] Delete the generated model


Source/ModelGroups/Animation_Skin.cs
```diff
@@ -352,63 +352,63 @@ Runtime.Animation CreateFoldingAnimation(Runtime.Node jointRootNode, List<Runtim
...
-                                Weight = weightToUse,
-                            });
-                        }
-
-                        weightList[weightIndex] = jointWeight;
-                    }
-
-                    properties.Add(new Property(PropertyName.Description, "`skinF`."));
-                }, (model) => { model.Camera = distantCamera; }),
+                //CreateModel((properties, animations, nodes) => {
```
bghgary: Would be good to put a comment before this saying why this is commented out.

donmccurdy:
> The spec isn't clear on this case yet...

I think the spec is fairly clear — most engines don't support >4, and the spec mentions that. Any particular details to add?

bghgary:
@stevk mentioned to me that the spec has this:

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#skinned-mesh-attributes
> The number of joints that influence one vertex is limited to 4, …

The issue with this is that we are saying we need a test model here: https://github.com/KhronosGroup/glTF/issues/1403

The >4 issue is here: https://github.com/KhronosGroup/glTF/pull/1381

The question for me is whether a model with >4 joints/weights per vertex is a positive test or a negative test.

Issue #475, opened 2018-12-12
Prep code for Jan 2019 Hackathon


stevk:
Below are the items that we want to have complete before the Hackathon, in descending priority order:

Tasks for @stevk :
- [x] Code style guide
- [x] Contributing guide
- [x] Create a ModelGroup class template
- [x] Update existing code to be compliant with the code style doc
- [x] #429 Add a Negative test flag to the manifest 
- [x] Test building the project in Visual Studio Code
- [x] #441 Make a model to handle asymmetric, non-zero tangents for cubic spline

Tasks for @kcoley :
- [x] #468 Change how the folder index is tracked so it doesn't change when model groups are inserted 
- [x] Test building the project in a Mac environment

Issue #448, opened 2018-09-05
Test needed for an asymetric non-zero tangents for cubic splines


kcoley:
Should resolve this issue: https://github.com/KhronosGroup/glTF-Asset-Generator/issues/441

Issue #441, opened 2018-07-31
asymmetric, non-zero tangents for cubic spline


andreasplesch:
Currently, all tangents for cubic spline interpolation in animations are zeroes:

https://github.com/bghgary/glTF-Asset-Generator/blob/master/Source/ModelGroups/Animation_Node.cs#L197

Consider using asymmetric tangents for a more demanding test asset. For example, the translation animation could involve the y axis to produce an interesting path.

andreasplesch:
Here is what I get with these 3 tuples of translation as output:

 0 0 0
-0.1 0 0
 0 0 0

 0 5 0 // tangent 5 in y
 0.1 0 0
 0 3 0 // tangent 3 in y

 0 0 0
 -0.1 0
 0 0 0 0

![nonzerotangents](https://user-images.githubusercontent.com/6171115/43467043-da5ad932-94ae-11e8-943d-ca7f3491e07c.gif)

bghgary:
> Consider using asymmetric tangents for a more demanding test asset.

Sounds good. Just out of curiosity, is there a specific situation where this can fall down? I'm trying to decide whether we need two tests or one. Please feel free to open a PR to make the change.

andreasplesch:
See #442 , unfortunately, I do not have a way to test the changes, including if other edits are required as a consequence.

I cannot really think of a situation where asymmetry would be catastrophic. Large tangents can have large effects, obviously, but this would not require an additional test.

kcoley:
Hi @andreasplesch , sorry we missed your PR!  We can make a model to handle this case for the Animation_Node tests.

stevk:
Fixed in PR #477

PR #477, opened 2018-12-21
Add a Animation_Node model with asymmetric tangents for cubic spline

stevk:
[Animation_Node Readme](https://github.com/stevk/glTF-Asset-Generator/blob/asymmetriccubicspline/Output/Animation_Node/README.md)

Fix for #441, #448 
Progress on #475


Source/ModelGroups/Animation_Node.cs
```diff
@@ -180,8 +180,23 @@ void SetStepSamplerForTranslation(List<Property> properties, Runtime.AnimationCh
...
+            void SetCubicSplineSamplerForTranslation(List<Property> properties, Runtime.AnimationChannel channel, bool useAsymmetricTangents = false)
             {
+                Vector3 inTangent;
+                Vector3 outTangent;
+                string propertyValueText = "Cubic Spline";
+            if (useAsymmetricTangents)
+                {
+                    inTangent = new Vector3(0.0f, 1.0f, 0.0f);
+                    outTangent = new Vector3(0.0f, -0.5f, 0.0f);
+                    propertyValueText = $"{propertyValueText} with <br>asymmetric tangents";
```
bghgary: I'm wondering if we need two different samples for translation. Why not just always use the asymmetric one?

stevk: I don't have a specific reason. I thought the decision back on Dec 12th was to add another model. Easy enough to combine them.

bghgary: I remember talking about this, but I'm not sure it's good since it introduces difficulty in explaining the idea with an md table.

PR #478, opened 2018-12-21
Add a negative test flag to the manifest

### LINKED ISSUE #429: Manifest.json should indicate if the asset should load or not.
In particular Compatibility_05.gltf shouldn't load; the json should have a field saying so.

stevk:
Added a bool to the manifest to show if a model is expected to successfully load. This value defaults to true.

Fixes #429

### DIFF
--- a/Output/Animation_Node/Manifest.json
+++ b/Output/Animation_Node/Manifest.json
@@ -3,7 +3,6 @@
   "models": [
     {
       "fileName": "Animation_Node_00.gltf",
-      "expectedToLoadSuccessfully": true,
       "sampleImageName": "Figures/SampleImages/Animation_Node_00.png",
       "camera": {
         "translation": [
@@ -15,7 +14,6 @@
     },
     {
       "fileName": "Animation_Node_01.gltf",
-      "expectedToLoadSuccessfully": true,
       "sampleImageName": "Figures/SampleImages/Animation_Node_01.png",
       "camera": {
         "translation": [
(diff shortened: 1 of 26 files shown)


Output/Animation_Node/Manifest.json
```diff
@@ -3,6 +3,7 @@
   "models": [
     {
       "fileName": "Animation_Node_00.gltf",
+      "expectedToLoadSuccessfully": true,
```
bghgary: Perhaps default this to true so that it doesn't serialize unless this value is false?

Source/Manifest.cs
```diff
@@ -19,13 +19,15 @@ public Manifest(ModelGroupId modelGroupId)
         public class Model
         {
             public string FileName;
+            public bool ExpectedToLoadSuccessfully;
             [JsonProperty( NullValueHandling = NullValueHandling.Ignore )]
             public string SampleImageName;
             public Camera Camera;
 
-            public Model(string name, ModelGroupId modelGroupId, bool noSampleImages, Camera cameraPositioning)
+            public Model(string name, ModelGroupId modelGroupId, bool noSampleImages, Camera cameraPositioning, bool willLoad)
```
bghgary: I'm not sure about the naming of `ExpectedToLoadSuccessfully`, but this part should at least match that name.

Output/Compatibility/Manifest.json
```diff
@@ -53,6 +53,7 @@
     },
     {
       "fileName": "Compatibility_05.gltf",
+      "expectedToLoadSuccessfully": false,
```
bghgary: We discussed offline.

Use `valid` as the property to mean whether the asset is a valid glTF. This is different than what the current flag means. It will only be set to false when an asset is not a valid glTF.

PR #482, opened 2019-01-09
Update primitive restart readme with invalid models notification

stevk:


Output/Mesh_PrimitiveRestart/README.md
```diff
@@ -1,4 +1,4 @@
-These models are intended to test behavior of primitive restart feature of graphics APIs.
+These models are intended to test behavior of primitive restart feature of graphics APIs. These are not valid glTF models as primitive restart values are disallowed. [See the glTF specification](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#primitiveindices) for more details.
```
bghgary: @lexaknyazev I noticed that the spec property reference has text that doesn't exist in the schema (the Minimum and Maximum). Did you add these by hand?

bghgary: Ah, I looked at that and for some reason still thought it was you. 😉

cc @donmccurdy

donmccurdy: @lexaknyazev what do you mean by rearranged? Sounds like we need to add the min/max in the schema?

donmccurdy:
I would consider making it more prominent that these are invalid glTF files.

Output/Mesh_PrimitiveRestart/README.md
lexaknyazev: Schema-defined fields refer to `primitive.indices` _property_, not accessor's content.

bghgary:
> I would consider making it more prominent that these are invalid glTF files.

Do you have a recommendation? Currently, the `manifest.json` has a flag that indicates invalid glTF assets. This PR adds wording that indicate this for this group.

Output/Mesh_PrimitiveRestart/README.md
donmccurdy: Ahhh ok yes, my mistake. 😅

donmccurdy:
Bold or block quote text would be enough I think. If they appear in the viewer some UI would be helpful there, too.

I could maybe see an argument for not keeping the invalid models in the repo, but having a flag to generate them as needed. Unclear what most users would do with these.

Output/Mesh_PrimitiveRestart/README.md
bghgary: Looks like we all overlooked where the text was being added when reviewing the PR. 😕

bghgary:
> Bold or block quote text would be enough I think.

👍 

> If they appear in the viewer some UI would be helpful there, too.

Perhaps the View links should be removed for these assets.

> I could maybe see an argument for not keeping the invalid models in the repo, but having a flag to generate them as needed. Unclear what most users would do with these.

Yeah, we've discussed this. We considered putting these in a completely separate folder.

Output/Mesh_PrimitiveRestart/README.md
donmccurdy: Opened https://github.com/KhronosGroup/glTF/pull/1532

PR #484, opened 2019-01-10
added id to manifest files

kcoley:
This PR would be dependent on an update on glTF-Assets-Viewer

We can discuss compared to this PR: https://github.com/KhronosGroup/glTF-Asset-Generator/pull/483

### DIFF
--- a/Source/ModelGroup.cs
+++ b/Source/ModelGroup.cs
@@ -225,29 +225,33 @@ namespace AssetGenerator
         public bool Valid = true;
     }
 
+    /// <summary>
+    /// The model group enum value associated with a model group should never change after a release
+    /// to avoid changes to the asset urls in the README files
+    /// </summary>
     internal enum ModelGroupId
     {
-        Animation_Node,
-        Animation_NodeMisc,
-        Animation_Skin,
-        Animation_SkinType,
-        Buffer_Interleaved,
-        Compatibility,
-        Material,
-        Material_AlphaBlend,
-        Material_AlphaMask,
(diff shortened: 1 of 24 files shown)


Source/ModelGroup.cs
```diff
@@ -211,26 +211,26 @@ internal class Model
...
-        Material_SpecularGlossiness,
-        Mesh_PrimitiveAttribute,
-        Mesh_PrimitiveMode,
-        Mesh_PrimitiveVertexColor,
-        Mesh_Primitives,
-        Mesh_PrimitivesUV,
-        Node_Attribute,
-        Node_NegativeScale,
-        Texture_Sampler,
+        Animation_Node = 0,
```
bghgary: It would be good to add a comment here saying these values should never change after a release.

Source/ModelGroup.cs
```diff
@@ -211,26 +211,26 @@ internal class Model
...
+        Material = 6,
+        Material_AlphaBlend = 7,
+        Material_AlphaMask = 8,
+        Material_DoubleSided = 9,
+        Material_MetallicRoughness = 10,
+        Material_Mixed = 11,
+        Material_SpecularGlossiness = 12,
+        Mesh_PrimitiveAttribute = 13,
+        Mesh_PrimitiveMode = 14,
+        Mesh_PrimitiveVertexColor = 15,
```
stevk: Missing Mesh_PrimitiveRestart. We talked about this offline.

stevk:
Required for #475

PR #488, opened 2019-01-15
Add missing primitive restart model group ID to manifest

stevk:
It looks like the build was not run after assigning the primitive restart model group an ID.


bghgary:
What created the build before that had the different ids?

bghgary:
I'm asking because the ids should no longer change.

bghgary:
We talked offline and this was an error when merging.

PR #492, opened 2019-01-18
Add method to better handle conflicts when writing files

### LINKED ISSUE #481: Sometimes files remain locked when running the build multiple times
Likely can be fixed by adding some kind of try/wait when writing files.

Error message in VS code `An unhandled exception of type 'System.IO.IOException' occurred in System.IO.FileSystem.dll: 'The requested operation cannot be performed on a file with a user-mapped section open'`

stevk:
Previously the app could crash when attempting to write a file if the file was already in use or otherwise locked. Now it retries a few times with a delay, and then halts the app and messages the user to address the issue.

Fixes #481

### DIFF
--- a/Source/FileHelper.cs
+++ b/Source/FileHelper.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 
 namespace AssetGenerator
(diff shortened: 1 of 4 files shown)


Source/FileHelper.cs
```diff
@@ -37,6 +37,35 @@ public static void ClearOldFiles(string executingAssemblyFolder, string assetFol
             }
         }
 
+        /// <summary>
```
bghgary: This shouldn't be necessary. I'm not sure why we are adding this.

stevk: I've seen Visual Code throw an error if the build is run more than once in a row (not a huge deal).
It also seemed prudent to have some kind of retry logic when a file fails to write for other reasons, like the file being in use from another app or not having write permissions to the location.

stevk: The auto retry is a fix for #481, where not being able to overwrite the file is a transient state. Otherwise the workaround for this issue is to run the build again. This is probably an issue that I'd encounter more than anyone else.

The manual retry covers resolving permission or file being in use by another app, with the assumption that the user will use the pause during the prompt to attempt a fix. I hadn't considered that this would be run as a part of automation, so it would be simpler to just not have this change at all in that regard.

bghgary: > The auto retry is a fix for #481, where not being able to overwrite the file is a transient state.

I don't think this is the right fix. We need to figure out why the process is not releasing the file when the process exits.

PR #493, opened 2019-01-22
Remove texture coords from models that don't have a texture

stevk:
Only add texture cords to models that have a texture. This fix affects a couple of the models in the  Material, Material_AlphaBlend, and Material_MetallicRoughness model groups. Before this fix models with texture cords but no texture have an error in the validator.

This wasn't a recent regression. These models have always had this issue, and I think the validator added this error check since the late time we validated all of the models.


Source/ModelGroups/Material.cs
```diff
@@ -23,13 +23,17 @@ public Material(List<string> imageList)
...
                 var properties = new List<Property>();
-                Runtime.MeshPrimitive meshPrimitive = MeshPrimitive.CreateSinglePlane();
-                meshPrimitive.Material = new Runtime.Material();
-                meshPrimitive.Material.MetallicRoughnessMaterial = new Runtime.PbrMetallicRoughness();
+                Runtime.MeshPrimitive meshPrimitive = MeshPrimitive.CreateSinglePlane(includeTextureCoords: false);
 
                 // Apply the common properties to the gltf.
-                meshPrimitive.Material.MetallicRoughnessMaterial.MetallicFactor = metallicFactorValue;
-                meshPrimitive.Material.MetallicRoughnessMaterial.BaseColorFactor = baseColorFactorValue;
+                meshPrimitive.Material = new Runtime.Material()
```
bghgary: nit: `()` can be omitted (multiple places)

PR #496, opened 2019-01-26
Add script for running the validator

### LINKED ISSUE #368: Validator needs to run across all models

stevk:
Added an npm script, which can be run as a launch configuration from VS Code.
The results are saved as raw output per model and an overall summary. These have their own folder structure as a subfolder in the Output folder.

[Summary README](https://github.com/stevk/glTF-Asset-Generator/blob/valid/ValidatorResults/README.md) 
[Individual results](https://github.com/stevk/glTF-Asset-Generator/blob/valid/ValidatorResults) 

Fixes #368


stevk:
Instead of using a powershell script I'm planning on validating models every time the build is run. Hold off on reviewing this PR until I implement that change.

Documentation/Adding_Models.md
```diff
@@ -106,6 +107,18 @@ At the bottom of the model group `GenerateUsedPropertiesList()` is called. This
 Screenshots are generated in a step separately from running the glTF Asset Generator, which also includes the moving of textures and figures into the output folders.
 This is done to speed up debugging. The creation of screenshots is a time intensive process and often the screenshots are not needed until the majority of debugging has been completed.
 
+## Validate Models
+1. [Install NPM](https://www.npmjs.com/get-npm)
+2. Open the [Validator](../Validator) folder in a command prompt and run the following commands:
+    ```
+    npm install --save babylonjs
```
bghgary: This is only necessary to create the package.json. You don't need to do this every time. Once the package.json is created, uses only need to run `npm install`

Output/ValidationResults.log
```diff
@@ -0,0 +1,299 @@
+﻿
```
bghgary: I'm not sure this file should be committed. Is there a benefit?

Documentation/Adding_Models.md
bghgary: Also, why do you need BabylonJS module?

ValidateModels.ps1
```diff
@@ -0,0 +1,10 @@
+$outputFolder = "Output"
```
bghgary: Since the validator comes from npm which is node, it probably makes the most sense to write this in JavaScript using node.

Validator/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "validate",
```
bghgary: not sure about the strings in this file. let's chat offline.

Documentation/Adding_Models.md
stevk: The `loadManifestFile` function in Validator/app/main.js calls a function from the BabylonJS module on Line 112. The `loadManifestFile` was reused from the screenshot generator code. There is likely another way to do the same thing without using the BabylonJS module.

```js
const rootDirectory = BABYLON.Tools.GetFolderPath(convertToURL(manifestJSON));
```

stevk: Now the documentation only instructs the user to run `npm install` while in that directory.

ValidateModels.ps1
stevk: I've added in similar changes as to what is in #497, so the npm script is run directly from C# instead of using powershell.

stevk:
Action Items
- [ ] Make an npm script to call the validator. Don't invoke from C#. 
- [ ] Setup a VS Code Launch.json way to run with validator
- [ ] Check all models, not only valid ones. 
- [ ] Documentation: Have a 'dependency' section above and separate from 'setup' 
Validator output changes 
    - [ ] Print the raw json
    - [ ] One output file per model. 
    - [ ] Mirror model directory in a folder in Output, and put the log files there. 
    - [ ] Make an all up report as well. Stats TBD but use the current totals for now as placeholder. 
    - [ ] MD table report. List of models. Everything under 'Issues' object, Link in report to that model. 
    - [ ] Check output file in for now

stevk:
There are still some outstanding issues 

- [ ] [Summary README](https://github.com/stevk/glTF-Asset-Generator/blob/valid/Output/_ValidatorResults/README.md) is too wide
Options are:
    - Reduce the number of columns.
    - Use smaller column names.
    - Use a model group section header to reduce the model name width)

- [ ] [Individual results](https://github.com/stevk/glTF-Asset-Generator/blob/valid/Output/_ValidatorResults/Animation_Node/Animation_Node_00.gltf.log) include a `validatedAt` field which is updated each time they are validated. This will add a lot of noise to PRs if we check in these files.
Options are:
    - Don't check these in.
    - Remove this info.

- [ ] Individual results don't have a newline at end of file.

Validator/app/main.js
```diff
@@ -0,0 +1,132 @@
...
+            status = ':white_check_mark:'
+        }
+
+        const modelLink = `[${glTFAssets[i].fileName.slice(0, -5)}](${convertToURL(path.join('..', glTFAssets[i].modelGroup, glTFAssets[i].fileName))})`;
+        const messages = JSON.stringify(issues.messages).split(',').join('<br>').split('}').join('<br>').split('{').join('').replace('[', '').replace(']', '');
+        summary.push(`| ${modelLink} | ${status} | ${issues.numErrors} | ${issues.numWarnings} | ${messages} |\n`);
+    }
+
+    // Write the summary report to file.
+    fs.writeFile(path.join(logOutputFolder, 'README.md'), summary.join('').split('\n').join('\r\n'), (err) => {
```
bghgary: don't fix line endings instead use gitattributes or something like it

Validator/app/main.js
```diff
@@ -0,0 +1,132 @@
...
+});
+
+/**
+ * Wraps the code for calling the Validator in a function so that it can be used as a promise.
+ * @param glTFAsset Object containing location and name metadata of a model. Created by createGLTFAsset().
+ */
+function validateModel(glTFAsset) {
+    const asset = fs.readFileSync(glTFAsset.filePath);
+
+    return validator.validateBytes(new Uint8Array(asset), {
```
bghgary: see [full usage](https://www.npmjs.com/package/gltf-validator#full-usage) to also validate the external files

Validator/app/main.js
```diff
@@ -0,0 +1,132 @@
...
+        uri: glTFAsset.fileName,
+    }).then((report) => {
+        glTFAsset.report = report;
+
+        //  Write the results to file.
+        const modelDirectory = path.join(logOutputFolder, path.basename(glTFAsset.fileDirectory));
+        delete report.validatedAt;
```
bghgary: comment this

Validator/app/main.js
```diff
@@ -0,0 +1,132 @@
...
+        try {
+            fs.mkdirSync(modelDirectory, { recursive: true } );
+        } catch (e) {
+            console.log('Cannot create folder ', e);
+        }
+        fs.writeFile(path.join(modelDirectory, glTFAsset.fileName) + '.log', (JSON.stringify(report, null, 4).split('\n').join('\r\n') + '\r\n'), (err) => {
```
bghgary: ditto about line endings

Validator/app/main.js
```diff
@@ -0,0 +1,132 @@
...
+            if (err) throw err;
+        });
+
+        // Write a simple result to console as models are completed.
+        if(parseInt(report.issues.numErrors) > 0) {
```
bghgary:
```suggestion
        if (parseInt(report.issues.numErrors) > 0) {
```

Validator/app/main.js
```diff
@@ -0,0 +1,132 @@
...
+ */
+function loadManifestFile(outputFolder, manifestJSON) {
+    const result = [];
+
+    // Open the manifest file.
+    const content = fs.readFileSync(manifestJSON);
+
+    // Create an object for each model's metadata.
+    const jsonData = JSON.parse(content);
+    for (let i = 0; i < jsonData.length; ++i) {
```
bghgary: ditto on looping

Validator/app/main.js
```diff
@@ -0,0 +1,132 @@
...
+ * @param manifestJSON Filepath to the manifest.
+ */
+function loadManifestFile(outputFolder, manifestJSON) {
+    const result = [];
+
+    // Open the manifest file.
+    const content = fs.readFileSync(manifestJSON);
+
+    // Create an object for each model's metadata.
+    const jsonData = JSON.parse(content);
```
bghgary:
```suggestion
    const modelGroups = JSON.parse(content);
```

Validator/app/main.js
```diff
@@ -0,0 +1,132 @@
...
+    return result;
+}
+
+/**
+ * Assembles location and name metadata of the glTF model.
+ * @param model JSON object of the model's metadata, extracted from the manifest.
+ * @param rootDirectory Directory that the model is located in.
+ * @param modelFolder Name of the folder containing the model. Is also the name of the modelgroup the model belongs to.
+ */
+function createGLTFAsset(model, rootDirectory, modelFolder) {
```
bghgary: consider inline this

Validator/app/main.js
```diff
@@ -0,0 +1,132 @@
...
+        const issues = glTFAssets[i].report.issues;
+        let status;
+        if (parseInt(issues.numErrors) > 0) {
+            status = ':x:';
+        }
+        else{
+            status = ':white_check_mark:'
+        }
+
+        const modelLink = `[${glTFAssets[i].fileName.slice(0, -5)}](${convertToURL(path.join('..', glTFAssets[i].modelGroup, glTFAssets[i].fileName))})`;
```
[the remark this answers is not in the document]
bghgary: delete convertToURL, don't use path.join, use `/` directly

stevk:
<br>
Validator Folder

- [ ] Change the name from '\app\main.js' to 'Validate.,js'
- [ ] Change the name from 'Validator' to 'Tools'
- [ ] Change command from 'Start' to 'Validate' for running the script

<br>
Package.json

- [ ] Remove all properties possible

<br>
Summary README

- [ ] Link to raw model report instead of the model
- [ ] Remove messages column
- [ ] Re-add Infos, Hints, Truncated
- [ ] Have each model group as a separate table

<br>
_ValidatorResults folder

- [ ] Move up one level to root
- [ ] Remove underscore

<br>
Documentation

- [ ] Update with path changes
- [ ] Update with 'npm run validate' change

<br>
Misc

- [ ] Run VS Code auto formatter
- [ ] Fix primitive restart issue by using full validator example call to validate the bin as well

ValidatorResults/README.md
```diff
@@ -0,0 +1,252 @@
+# Animation  Node
+| Model | Status | Errors | Warnings | Infos | Hints | Truncated |
```
bghgary: Remove truncated.

PR #497, opened 2019-01-31
Remove powershell dependency by integrating SampleImageHelper.ps1

### LINKED ISSUE #487: Find a better alternative for SampleImageHelper.ps1
**Options**
- Convert the powershell script into C#. Run with a switch of some kind.
- Python script?

stevk:
Converted the script into C#. Now screenshots will be generated when the app is passed the arg `GenerateScreenshots`

Fixes #487


Source/npmHelper.cs
```diff
@@ -0,0 +1,126 @@
...
+            Directory.CreateDirectory(tempOutputDirectory);
+            Directory.CreateDirectory(tempSampleImagesDirectory);
+            Directory.CreateDirectory(tempThumbnailsDirectory);
+
+            // Runs the ScreenshotGenerator.
+            Console.WriteLine();
+            Console.WriteLine("Running the ScreenshotGenerator");
+            var psiNpmStartScreenshotGenerator = new ProcessStartInfo
+            {
+                FileName = "cmd",
```
stevk: Invoking cmd here. Pretty sure this won't work on IOS, so I'll need to rethink this approach.

stevk: Invoking cmd here. Pretty sure this won't work on IOS, so I'll need to rethink this approach.

Source/npmHelper.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            var pathsArray = Environment.GetEnvironmentVariable("Path").Split(';');
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return Path.Combine(pathsArray.First(x => x.Contains("npm")), "npm.cmd");
+            }
+            else
+            {
+                // Should be /usr/local/bin/npm by default
+                return pathsArray.First(x => x.Contains("npm"));
```
stevk: I think this will work for OSX and Linux, but I need to test it to be sure. Pending my getting a hold of a mac...

PR #500, opened 2019-03-06
Add script for generating screenshots and remove powershell dependancy

### LINKED ISSUE #487: Find a better alternative for SampleImageHelper.ps1

stevk:
Screenshots are now generated by using the `Generate Screenshots` launch configuration, or by using `npm run generateScreenshots` from the tools directory. This removes our dependency on powershell.

Note that the GenerateScreenshots repro is still external to the project and still needs to be setup separately, the same as before.

Fixes #487

### DIFF
--- a/Tools/generateScreenshots.js
+++ b/Tools/generateScreenshots.js
@@ -1,4 +1,3 @@
-const os = require('os');
 const fs = require('fs');
 const path = require('path');
 const exec = require('child_process').exec;
@@ -17,13 +16,6 @@ const defaultImage = path.join(resourcesDirectory, "Figures", "NYI.png");
 const screenshotGeneratorRoot = path.join(projectDirectory, "ScreenshotGenerator");
 const screenshotGeneratorDirectory = path.join(screenshotGeneratorRoot, 'app');
 const resizeScript = path.join(screenshotGeneratorRoot, "pythonScripts", "dist", "resizeImages.exe");
-
-// Load the manifest into memory.
-var manifest;
-fs.readFile(manifestPath, function(error, data) {
-    if (error) throw error;
-    manifest = JSON.parse(data);
- });
  
 // Create some temp folders. The screenshots will be stored here after generation until being sorted into the Output folder.
 try {
@@ -33,6 +25,13 @@ try {
 } catch (e) {
     console.log('Cannot create folder ', e);
 }
+
+// Load the manifest into memory.
+var manifest;
+fs.readFile(manifestPath, function(error, data) {
+    if (error) throw error;
+    manifest = JSON.parse(data);
+ });
 
 // Generate the screenshots via the ScreenshotGenerator.
 const callScreenshotGenerator = `npm start -- "headless=true" "manifest=${manifestPath}" "outputDirectory=${tempSampleImagesDirectory}"`;
@@ -46,9 +45,9 @@ runProgram(callScreenshotGenerator, screenshotGeneratorDirectory, function() {
     runProgram(callThumbnailGenerator, screenshotGeneratorDirectory, function() {
         console.log('Finished generating thumbnails.');
         console.log('');
+        console.log('Copying images to the Output directory...');
         sortImages()
         .then(
-            // Delete the temp images directory.
             console.log('Finished copying images to the Output directory.'),
             console.log(''),
             console.log('Cleaning up temporary files...'),
@@ -58,6 +57,9 @@ runProgram(callScreenshotGenerator, screenshotGeneratorDirectory, function() {
     });
 });
 
+/**
+ * Moves screenshots and thumbnails from the temp folders into their final location in the Output directory.
+ */
 function sortImages() {
     return new Promise(resolve => {
         // Check for images that are pre-generated.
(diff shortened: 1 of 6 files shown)


Tools/generateScreenshots.js
```diff
@@ -0,0 +1,174 @@
...
+try {
+    fs.mkdirSync(tempOutputDirectory, { recursive: true } );
+    fs.mkdirSync(tempSampleImagesDirectory, { recursive: true } );
+    fs.mkdirSync(tempThumbnailsDirectory, { recursive: true } );
+} catch (e) {
+    console.log('Cannot create folder ', e);
+}
+
+// Load the manifest into memory.
+var manifest;
```
bghgary: Move this section to below so that it's localized to the code that uses it.

Tools/generateScreenshots.js
```diff
@@ -0,0 +1,174 @@
...
+fs.readFile(manifestPath, function(error, data) {
+    if (error) throw error;
+    manifest = JSON.parse(data);
+ });
+
+// Generate the screenshots via the ScreenshotGenerator.
+const callScreenshotGenerator = `npm start -- "headless=true" "manifest=${manifestPath}" "outputDirectory=${tempSampleImagesDirectory}"`;
+const callThumbnailGenerator = `${resizeScript} --dir=${tempSampleImagesDirectory} --outputDir=${tempThumbnailsDirectory} --width=72 --height=72`;
```
bghgary: put quotes around arguments that have paths

Tools/generateScreenshots.js
```diff
@@ -0,0 +1,174 @@
...
+console.log();
+console.log('Running the ScreenshotGenerator...');
+runProgram(callScreenshotGenerator, screenshotGeneratorDirectory, function() {
+    console.log('Finished generating screenshots.');
+    console.log('');
```
bghgary: remove empty string

Tools/generateScreenshots.js
bghgary: Move to right next to where you use it

Tools/generateScreenshots.js
```diff
@@ -0,0 +1,174 @@
...
+    console.log('Creating thumbnails...');
+    runProgram(callThumbnailGenerator, screenshotGeneratorDirectory, function() {
+        console.log('Finished generating thumbnails.');
+        console.log('');
+        console.log('Copying images to the Output directory...');
+        sortImages()
+        .then(
```
bghgary: should be this:
```
.then(function () {
   // ...
});
```

Tools/generateScreenshots.js
```diff
@@ -0,0 +1,174 @@
...
+            console.log('Clean up complete.'),
+        );
+    });
+});
+
+/**
+ * Moves screenshots and thumbnails from the temp folders into their final location in the Output directory.
+ */
+function sortImages() {
+    return new Promise(resolve => {
```
bghgary: make this function synchronous

bghgary: just make everything synchronous

Tools/generateScreenshots.js
```diff
@@ -0,0 +1,174 @@
...
+const screenshotGeneratorDirectory = path.join(screenshotGeneratorRoot, 'app');
+const resizeScript = path.join(screenshotGeneratorRoot, "pythonScripts", "dist", "resizeImages.exe");
+ 
+// Create some temp folders. The screenshots will be stored here after generation until being sorted into the Output folder.
+try {
+    fs.mkdirSync(tempOutputDirectory, { recursive: true } );
+    fs.mkdirSync(tempSampleImagesDirectory, { recursive: true } );
+    fs.mkdirSync(tempThumbnailsDirectory, { recursive: true } );
+} catch (e) {
+    console.log('Cannot create folder ', e);
```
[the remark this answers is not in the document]
stevk: It does format right, but the code continues. So I've added a `throw error`.

Tools/generateScreenshots.js
stevk: Turns out that `''` is needed to get it to write an empty line. Updated all of the calls to have this.

Tools/generateScreenshots.js
stevk: Like I mentioned offline, I've not been able to get execSync to pipe live output into the console. I can save and dump the output after the process has finished, but doing this leave the app looking like it is hung for a long time while screenshots are created.

Tools/generateScreenshots.js
```diff
@@ -6,129 +6,119 @@ const exec = require('child_process').exec;
...
+const tempSampleImagesDirectory = path.join(tempOutputDirectory, 'SampleImages');
+const tempThumbnailsDirectory = path.join(tempOutputDirectory, 'Thumbnails');
+const defaultImage = path.join(resourcesDirectory, 'Figures', 'NYI.png');
+const screenshotGeneratorRoot = path.join(projectDirectory, 'ScreenshotGenerator');
 const screenshotGeneratorDirectory = path.join(screenshotGeneratorRoot, 'app');
-const resizeScript = path.join(screenshotGeneratorRoot, "pythonScripts", "dist", "resizeImages.exe");
+const resizeScript = path.join(screenshotGeneratorRoot, 'pythonScripts', 'dist', 'resizeImages.exe');
  
 // Create some temp folders. The screenshots will be stored here after generation until being sorted into the Output folder.
 try {
```
bghgary: consider removing the try/catch completely

Tools/generateScreenshots.js
```diff
@@ -6,129 +6,119 @@ const exec = require('child_process').exec;
...
-                    return true;
-                } else return false;
-            })
-        });
-        if (existingImageList.length > 0) {
-            console.log('The following generated image(s) will not be used, due to there already being a pre-generated image:');
-            for (const image in existingImageList) {
-                console.log(image);
-            }
+    genImages.forEach(function (genImage) {
```
bghgary: use `for (... of ...) syntax` with a break

Tools/generateScreenshots.js
```diff
@@ -6,129 +6,119 @@ const exec = require('child_process').exec;
...
-                    if (existingImageList.Count > 0) {
-                        for (const preGenImage in existingImageList) {
-                            if (existingImageList[preGenImage] == path.basename(model.sampleImageName)) {
-                                imageSource = path.join(resourcesDirectory, model.sampleImageName);
-                                imageThumbnailSource = path.join(resourcesDirectory, thumbnailName);
-                                break;
-                            }
+                // Check if there is an pre-gen image, and use that filepath instead if it does.
+                if (existingImageList.Count > 0) {
+                    for (const preGenImage in existingImageList) {
```
bghgary: should be for of

PR #508, opened 2019-03-14
Split the Output folder into Positive and Negative tests

### LINKED ISSUE #485: Invalid models should be better seperated from valid models
Options for separating invalid models
- Bold text with :warning: as readme header declaring the models as invalid.
- Secondary output folder for all invalid models.
- Invalid models don't have 'view' links.
- Invalid models aren't pre-generated. Requires switch to generate as needed.

Continuation of the conversation in #482

stevk:
⚠️ This is a breaking change! 
- Any previously created external links to models that reference master instead of an earlier blob will break.
- Links to view models will break until the glTF-Assets-Viewer is updated to handle the new manifest locations. (fix in bghgary/glTF-Assets-Viewer#11)

List of changes:
- Add two sub folders to Output, Positive and Negative. 
    - Each of these folders has its own manifest and validator results summary. 
    - The contents of both folders are listed separately in the main readme.
    - Update the screenshot generator and validator scripts to handle the new output paths.
- Add a short explanation of Positive vs Negative to the Output folder.
- Save raw validator results in a folder with the relevant model group.
- Stop marking models as invalid in the manifest (affects only Primitive Restart).

Fixes #485


Tools/generateScreenshots.js
```diff
@@ -139,7 +162,7 @@ function deleteFolderRecursive(folderPath) {
     var files = [];
     if (fs.existsSync(folderPath)) {
         files = fs.readdirSync(folderPath);
-        files.forEach(function(file,index) {
+        files.forEach((file, index) => {
```
bghgary: nit: index isn't used

PR #509, opened 2019-03-15
Tag models as loadable or not

### LINKED ISSUE #429: Manifest.json should indicate if the asset should load or not.

stevk:
Tag all Negative test models in the manifest as not loadable as well as Compatibility 04 and 05.

Fixes #429

### DIFF
--- a/Source/ModelGroups/Mesh_PrimitiveRestart.cs
+++ b/Source/ModelGroups/Mesh_PrimitiveRestart.cs
@@ -53,7 +53,7 @@ namespace AssetGenerator
                             }
                         }
                     }),
-                    Valid = false
+                    Loadable = false
                 };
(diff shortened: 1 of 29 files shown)


Source/ModelGroups/Mesh_PrimitiveRestart.cs
```diff
@@ -53,7 +53,7 @@ Model CreateModel(Action<List<Runtime.MeshPrimitive>> setProperties)
                             }
                         }
                     }),
-                    Valid = false
+                    Loadable = false
```
bghgary: Make loadable a nullable and should only be null for negative tests.

PR #523, opened 2019-05-14
Add a skinning edge case for multiple nodes pointing to the same mesh with different skins

### LINKED ISSUE #521: Add a skinning edge case for multiple nodes pointing to the same mesh with different skin
See model from https://github.com/donmccurdy/three-gltf-viewer/issues/147.

stevk:
This new model is done in a similar style as Animation_Skin_05, where there are two instances of skin A side by side. 

[Readme](https://github.com/stevk/glTF-Asset-Generator/blob/skinEdgeCase/Output/Positive/Animation_Skin/README.md)

[View Model](https://bghgary.github.io/glTF-Assets-Viewer/?manifest=https://raw.githubusercontent.com/stevk/glTF-Asset-Generator/skinEdgeCase/Output/Positive/Manifest.json&folder=2&model=11)

Fixes #521


stevk:
I would have preferred to color one of the skins blue, like with Animation_Skin_05, but both models are using the same mesh. I could give them unique animations if we don't mind the added complexity.

vpenades:
When I was testing the limits of 3D Visor, I noticed that somehow, it was able to handle up to two instances, the issue become apparent when I added many more instances, so I would suggest to add at least three instances.

bghgary:
@vpenades 
> When I was testing the limits of 3D Visor, I noticed that somehow, it was able to handle up to two instances, the issue become apparent when I added many more instances, so I would suggest to add at least three instances.

Do you know what is the root cause of this in 3D Visor?

Source/Runtime/JointWeight.cs
```diff
@@ -2,7 +2,7 @@
 {
     internal class JointWeight
     {
-        public SkinJoint Joint = null;
+        public int JointIndex = 0;
```
bghgary: consider initializing to -1

vpenades:
@bghgary 

> Do you know what is the root cause of this in 3D Visor?

When I try InfiniteSkinnedTentacle.glb with 3D View, only two instances of the mesh appear on screen; interestingly, one of the instances is animated whereas the other one is unanimated.

My guess is that View 3D assumes that models will come only with a single skin per scene, and the drawing pipeline only picks the first skinned mesh. The second, unanimated instance, I have no idea, thought.

Another possibility is that for the first mesh-skin instance, the skin transforms are set correctly, but for the subsequent skins, the transforms are messed, nullified, or reset, causing the rest of instances to render in odd ways, or not at all.

Without looking at the code is hard to say...

bghgary:
Lol, sorry, I keep forgetting 3D Visor is 3D Viewer

bghgary:
@stevk Can you try the new model in 3D viewer and check the result?

stevk:
@bghgary Model as seen in 3D Viewer.
image

bghgary:
@vpenades Looks like our test model already breaks the 3D viewer, so I don't think we need to create more than 2. Let me know what you think.

vpenades:
@bghgary it looks fine to me.

Two more things: I might take into account the skeleton referente in the skins... when I removed the skeleton reference the file was still valid, but it broke Cesium3D.

Also, in my example, I made one of the skin's bone tree to be the child of the other skin's tree, just to in case renderers assume skins are science robot based.

Maybe these two cases would maker more generated examples?

bghgary:
> when I removed the skeleton reference the file was still valid, but it broke Cesium3D.

That's a different test. We can do this separately. Mind filing an issue?

> one of the skin's bone tree to be the child of the other skin's tree, just in case renderers assume skins are scene root based.

I'm not sure I understand this one. Can you show me your example?

stevk:
> Also, in my example, I made one of the skin's bone tree to be the child of the other skin's tree, just in case renderers assume skins are scene root based.

@bghgaryI think @vpenades is describing a case is similar to [Animation_Skin_02](https://github.com/KhronosGroup/glTF-Asset-Generator/blob/master/Output/Positive/Animation_Skin/Animation_Skin_02.gltf
) where the nodes used by the skin are not root nodes. Two differences are that this existing model has only one skin instead of two skins on the same node heirarchy, and it doesn't have an animation (just a node rotation).

vpenades:
> @bghgary I think @vpenades is describing a case is similar to [Animation_Skin_02](https://github.com/KhronosGroup/glTF-Asset-Generator/blob/master/Output/Positive/Animation_Skin/Animation_Skin_02.gltf) where the nodes used by the skin are not root nodes. Two differences are that this existing model has only one skin instead of two skins on the same node heirarchy, and it doesn't have an animation (just a node rotation).

Exactly, furthermore, in my example, the mesh is the same!

So you have 1 mesh, instantiated 2 times with 2 skins pointing to the same node hierarchy, but to different joints in the hierarchy.

I think it's pretty wicked, but not completely out of order; Imagine you want to generate a tree (as in SpeedTree's forest trees) where a single leaf or brach is instantiated multiple times around the trunk, and you want the wind to move it gently with an animation...

bghgary:
Okay, sounds reasonable. Please file a separate issue for both. Thanks!

PR #525, opened 2019-05-16
Add a Material_DoubleSided model with a texture but no normals or tangents

### LINKED ISSUE #499: Material_DoubleSided group is missing a case
Need a test for no normals, no tangents, with normal texture.

stevk:
[Material_DoubleSided](https://github.com/stevk/glTF-Asset-Generator/blob/doublesidedModel/Output/Positive/Material_DoubleSided/README.md)

Fixes #499


stevk:
> Can we verify that Babylon is doing it correctly?

The new model's appearance is in line with the other doublesided test models, so it appears correct to me.

PR #526, opened 2019-05-16
Add Animation_Misc model that has ignored channels

### LINKED ISSUE #453: Animation with ignored channels
A suggested addition: an animation with two channels, where one channel omits `node`. Per spec:

> When node isn't defined, channel should be ignored. Valid path names are "translation", "rotation", "scale", and "weights".

... an implementation should ignore one channel and be able to play the other.

/cc https://github.com/KhronosGroup/glTF/pull/1301#issuecomment-422057805

stevk:
Adds a model with two channels. The first channel doesn't target a node and therefore, per the spec, should be ignored. The second channel should play as normal.

[Animation_NodeMisc](https://github.com/stevk/glTF-Asset-Generator/blob/ignoredChannel/Output/Positive/Animation_NodeMisc/README.md)

Fixes #453


Source/Runtime/GLTFConverter.cs
```diff
@@ -1243,28 +1243,12 @@ private Loader.Animation ConvertAnimationToSchema(Animation runtimeAnimation, Lo
...
-                var inputAccessor = CreateAccessor(bufferViews.Count - 1, 0, ComponentTypeEnum.FLOAT, runtimeSampler.InputKeys.Count(), "Animation Sampler Input", TypeEnum.SCALAR, null, max, min);
-                accessors.Add(inputAccessor);
-
-                int inputAccessorIndex = accessors.Count - 1;
-
+                // Create Animation Channel.
                 animationChannel.Target = new Loader.AnimationChannelTarget
                 {
-                    Node = targetNodeIndex
+                    Node = (targetNode != null) ? (int?)ConvertNodeToSchema(targetNode, gltf, buffer, geometryData, bufferIndex) : null
```
bghgary: might be clearer if you put this as a separate statement after creating the `Target`

```
if (targetNode != null)
{
    animationChannel.Target.Node = ...
}
```

PR #527, opened 2019-05-17
Add test models that don't use the position attribute

### LINKED ISSUE #407: Add assets with no POSITION attribute
One can argue this is a negative test, but since https://github.com/KhronosGroup/glTF/issues/1332 is asking for this, we can make an exception similar to the compatibility tests.

stevk:
[Mesh_NoPosition readme](https://github.com/stevk/glTF-Asset-Generator/tree/noPositions/Output/Negative/Mesh_NoPosition)

Fixes #407


Source/Property.cs
```diff
@@ -106,6 +106,8 @@ static object Process(object obj)
 
     internal enum PropertyName
     {
+        Indices,
+        Position,
```
bghgary:
```suggestion
        VertexPosition,
```

Source/Property.cs
```diff
@@ -106,6 +106,8 @@ static object Process(object obj)
 
     internal enum PropertyName
     {
+        Indices,
```
bghgary: How is this different than IndicesValues?

stevk: There isn't a difference. I'll switch to IndicesValues.

Source/Property.cs
```diff
@@ -106,8 +106,7 @@ static object Process(object obj)
 
     internal enum PropertyName
     {
-        Indices,
-        Position,
+        VertexPosition,
```
bghgary: Move together with the other Vertex properties.

PR #534, opened 2019-06-05
Add the Animation_SamplerType model group

### LINKED ISSUE #533: Need test models for different component types for animations
We'll cover these cases for now. 

[Animations Spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#animations)

"rotation" "VEC4" 
5120 (BYTE) normalized
5121 (UNSIGNED_BYTE) normalized
5122 (SHORT) normalized
5123 (UNSIGNED_SHORT) normalized 

Source issue: https://github.com/BabylonJS/Babylon.js/issues/6436

stevk:
Test models for animation sampler component types.
[Animation_SamplerType](https://github.com/stevk/glTF-Asset-Generator/tree/AnimationSamplerType/Output/Positive/Animation_SamplerType)

Fixes #533


Source/ModelGroups/Animation_SamplerType.cs
```diff
@@ -0,0 +1,127 @@
...
+                }),
+                CreateModel((properties, sampler) => {
+                    sampler.OutputComponentType = AnimationSampler.ComponentTypeEnum.NORMALIZED_BYTE;
+                    properties.Add(new Property(PropertyName.SamplerComponentType, "Byte"));
+                }),
+                CreateModel((properties, sampler) => {
+                    sampler.OutputComponentType = AnimationSampler.ComponentTypeEnum.NORMALIZED_SHORT;
+                    properties.Add(new Property(PropertyName.SamplerComponentType, "Short"));
+                }),
+                // Commenting these models out for now as rotations using unsigned types must be positive, which doesn't seem like a real world case.
```
bghgary: Recommend just remove it

Source/Property.cs
```diff
@@ -161,6 +161,7 @@ internal enum PropertyName
         ModelShouldLoad,
         JointsComponentType,
         WeightComponentType,
+        SamplerComponentType,
```
bghgary:
```suggestion
        SamplerOutputComponentType,
```

Source/Runtime/AnimationSampler.cs
```diff
@@ -5,6 +5,8 @@ namespace AssetGenerator.Runtime
     internal abstract class AnimationSampler
     {
         public abstract IEnumerable<float> InputKeys { get; }
+        public enum ComponentTypeEnum { FLOAT, NORMALIZED_BYTE, NORMALIZED_UNSIGNED_BYTE, NORMALIZED_SHORT, NORMALIZED_UNSIGNED_SHORT };
+        public ComponentTypeEnum OutputComponentType { get; set; }
```
bghgary: talked offline. this can be protected set and change the code above to pass in the output component type to CreateModel.

stevk:
@bghgary We'll need a npm babylonjs package with this fix before we can generate reference screenshots for these models.

bghgary:
@lexaknyazev For normalized non-float quaternions, it's impossible to get a perfectly unit quaternion. The validator is giving errors for `BYTE` (though not for `SHORT`), but I don't think we can make it any closer. See https://github.com/KhronosGroup/glTF-Asset-Generator/blob/34047f054151b3ec7cb49e217da3c8e4d1c31ed3/Output/Positive/Animation_SamplerType/ValidatorResults/Animation_SamplerType_01.json#L10. Also, should implementations normalize quaternions when the type is normalized non-float?

lexaknyazev:
> For normalized non-float quaternions, it's impossible to get a perfectly unit quaternion

I think there was a proposal to adjust the validation threshold so that it allows the error from ubyte quaternions.

PR #542, opened 2019-08-05
Refactor ReadmeStringHelper into ReadmeExtensionMehtods

stevk:
Refactors ReadmeStringHelper.cs into ReadmeExtensionMethods.cs, which implements extension methods for getting values as strings ready formatted for the readme tables. 
These extension methods are called via ToReadmeString() where needed.


Source/ReadmeExtensionMethods.cs
```diff
@@ -0,0 +1,194 @@
...
+using static AssetGenerator.Runtime.MeshPrimitive;
+
+namespace AssetGenerator
+{
+    /// <summary>
+    /// Adds an alternative to the ToString method on several values. Intended for use in printing nicely formatted human readable values.
+    /// </summary>
+    internal static class ReadmeExtensionMethods
+    {
+        internal static string ToReadmeString(this Vector3 value)
```
bghgary: no need to use internal for a function inside an internal class.. just use public

PR #553, opened 2019-10-31
Add Buffer_Misc model group to separate buffers

### LINKED ISSUE #440: individual buffers for POSITION, INDEX, NORMAL, animation channels
Perhaps add generation of individual buffers for POSITION, INDEX, NORMAL, animation channels etc. Some engines may expect that there is only one buffer, or one for geometry and one for animation.

YuuuZhang:
Intended to separate buffers into geometry data and animation data.

Fixes #440


Source/Program.cs
```diff
@@ -57,6 +57,7 @@ private static void Main(string[] args)
                 new Texture_Sampler(imageList),
                 new Animation_SamplerType(imageList),
                 new Instancing(imageList),
+                new Buffer_Misc(imageList),
```
bghgary: This list is supposed to be sorted. Please put Buffer_Misc in the right place. Then in a separate PR fix the last two entries so that they are also correct.

Source/ModelGroups/Buffer_Misc.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace AssetGenerator.ModelGroups
+{
+    internal class Buffer_Misc : ModelGroup
+    {
+        public override ModelGroupId Id => ModelGroupId.Buffer_Misc;
+
+        public Buffer_Misc(List<string> imageList)
+        {
+            NoSampleImages = true;
```
bghgary: Why no sample images for this group? Seems like we should have it.

Source/Program.cs
YuuuZhang: Got it. I`ll make Buffer_Misc in the right place. 
And correct `Animation_SamplerType` and `Instancing` in another PR.

Source/ModelGroups/Buffer_Misc.cs
YuuuZhang: Because sample images do not affect separate buffers and animation display effect.
So I do not use sample images to make the geometry settings to be minimal. 
But we can also add sample images to make the display effect more pretty.
[PS: Below is the preview of Buffer_Misc_00.gltf:
image ]

Source/ModelGroups/Buffer_Misc.cs
```diff
@@ -11,14 +11,23 @@ internal class Buffer_Misc : ModelGroup
 
         public Buffer_Misc(List<string> imageList)
         {
-            NoSampleImages = true;
+            UseFigure(imageList, "NYI");
```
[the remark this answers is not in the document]
bghgary: I removed this.

Source/ModelGroups/Buffer_Misc.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            // Track the common properties for use in the readme.
+            var baseColorFactorGreen = new Vector4(0.0f, 1.0f, 0.0f, 1.0f);
+            CommonProperties.Add(new Property(PropertyName.BaseColorFactor, baseColorFactorGreen.ToReadmeString()));
+
+            Model CreateModel(Action<List<Property>, AnimationChannel, Node> setProperties) 
+            {
+                var properties = new List<Property>();
+                Runtime.MeshPrimitive meshPrimitive = MeshPrimitive.CreateSinglePlane(includeTextureCoords: false);
+                meshPrimitive.Material = new Runtime.Material
```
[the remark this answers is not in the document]
bghgary: I also removed this.

Issue #550, opened 2019-09-26
Add test for instanced mesh with negative scale


bghgary:
If two nodes point to the same mesh and one of the nodes has a negative scale, engines must use two draw calls to render them to account for the negative determinant and flipping of winding order.

bghgary:
Fixed in #555

PR #555, opened 2019-11-08
Add negative scale to node.

YuuuZhang:
For issue #550 
- Include screen shots and validate results. 

- Add the negative scale model to position 07. 
[Together with 'Two nodes use same mesh(with different translations) ']


Source/ModelGroups/Instancing.cs
```diff
@@ -357,6 +357,20 @@ void AddAnimation(List<Animation> animations, List<Node> nodes, AnimationSampler
...
                 CreateModel((properties, nodes, animations) =>
+                {
+                    var meshPrimitive = MeshPrimitive.CreateSinglePlane();
+                    meshPrimitive.Material = CreateMaterial(CreateTextureInfo(baseColorImageA));
+
+                    AddMeshPrimitivesToMultipleNodes(nodes, meshPrimitive, meshPrimitive);
+                    nodes[1].Mesh = nodes[0].Mesh;
+
+                    // Adds just the node containing the negative scale.
+                    nodes[0].Scale = new Vector3(-1.0f, -1.0f, 1.0f);
```
bghgary: Flipping two axes doesn't test properly. The transform matrix will end up with a positive determinant which isn't going to test this case properly. We just need to flip one axis.

```suggestion
                    nodes[0].Scale = new Vector3(1.0f, -1.0f, 1.0f);
```

Source/ModelGroups/Instancing.cs
```diff
@@ -357,6 +357,20 @@ void AddAnimation(List<Animation> animations, List<Node> nodes, AnimationSampler
...
+
+                    properties.Add(new Property(PropertyName.Description, "Two nodes using the same mesh."));
+                    properties.Add(new Property(PropertyName.Difference, "One node has a negative scale and the other does not have a negative scale."));
```
bghgary:
```suggestion
                    properties.Add(new Property(PropertyName.Difference, "One node has a [1.0, -1.0, 1.0] scale and the other has default scale."));
```

Source/ModelGroups/Instancing.cs
YuuuZhang: Got it. Thanks Gary.

YuuuZhang:
> The sample image for the new model is now wrong. Can you investigate?

Yes, I`ll check the reason.

YuuuZhang:
> The sample image for the new model is now wrong. Can you investigate?

The new model image in _sandbox.babylonjs_ is right (show two 'BaseColor_A'pics). 
But after _GenerateScreenshots_ it goes wrong (only show one 'BaseColor_A' pic). 

I am still investigating the issue, btw, it is interesting that if I change node[0] scale (**1.0f**, -1.0f, 1.0f) to (**-1.0f**, -1.0f, 1.0f), the screenshots can show same images with _sandbox.babylonjs_  (two 'BaseColor_A'pics).

bghgary:
I think this was a bug in Babylon.js at some point. The model works now in the sandbox. You probably just need to update Babylon.js to the latest version for the screenshot.

PR #559, opened 2020-01-22
Fix unused joint indices

lexaknyazev:
This PR fixes `ACCESSOR_JOINTS_USED_ZERO_WEIGHT` warning for `Animation_Skin_07` and `Animation_Skin_10`.

Similar code in `CreateJointsAndWeightsForMultipleRoots` was not edited because it's currently unused.


lexaknyazev:
Yes, the validator follows this statement from the spec:
> Unused joint values (i.e. joints with a weight of zero) should be set to zero.

The reasoning is:
1. There could be cases when a vertex has less than 4 influences.
2. Yet, the accessor is `VEC4`, so something must be written to missing joint indices.
3. Given that we decided to not introduce a "special" index value (like 255 or 65535) for unused influences, there's no way to distinguish between truly unused (as if there are only 3 joints total) and "normal" zero-weight (like in this example) influences.
4. Therefore, all zero-weight influences should have zero joint index.

bghgary:
Yeah, I vaguely recall this discussion now. Thanks!

